PR #10, opened 2024-01-09
Declare SCION AS numbers independant from BGP ASNs.

### LINKED ISSUE #9: Adapt Section on SCION AS Numbers
(Moved from https://github.com/scionassociation/standards/issues/91)

Section 1.5.2 on AS numbering in the Control Plane draft must be changed, as it may cause problems or even a blocking of the draft at the IETF.  See also Eliot Lear's mail from 28.8:

> There is an issue in Section 1.5.2:
> There will be a conflict if the IETF decides to increase the ASN size.  This may not make it past the IESG in RFC 5742 review.  I just went through this with GNS and would not like to repeat that experience.  Therefore, my suggestion is that you view the ASN field as two separate fields: BASE and and EXTENSION, where if EXTENSION is 0, base must be an IANA-assigned ASN value.
> 

At the 3rd TC meeting on Tuesday 5 September, it was decided that the SCION AS numbering will not be described in detail in the CP draft. It should be mentioned that the SCION AS numbering is totally independent from the BGP AS numbering. For all other details, refer to a future and/or other draft.   

For the problem description and possible future solution, see slide 11-14 of the slide deck for the TC Standardisation meeting on 5.9.23. [TC-Standardisation_3meeting_Sept23.pdf](https://github.com/scionassociation/standards/files/12534358/TC-Standardisation_3meeting_Sept23.pdf)

jiceatscion:
This is my proposal to address the scion/BGP AS number objection.

Fixes #9

### DIFF
--- a/draft-dekater-scion-controlplane.md
+++ b/draft-dekater-scion-controlplane.md
@@ -323,7 +323,7 @@ Currently, ISD numbers are allocated by Anapaya, the Swiss-based provider of SCI
 
 ### AS Numbers
 
-An AS number is the 48-bit identifier for an AS.  SCION inherits the existing 32-bit AS numbers from RFC4893, but provides an extended 48-bit space, allowing for additional SCION-only AS numbers beyond the 32-bit space in use today.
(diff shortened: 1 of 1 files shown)


nicorusti (approved):
I edited a bit of wording, it overall looks good to me.

PR #11, opened 2024-01-11
Improve references classification and uses of MUST and SHOULD

jiceatscion:
There's still some more MUST required IMHO.

### DIFF
--- a/draft-dekater-scion-controlplane.md
+++ b/draft-dekater-scion-controlplane.md
@@ -114,19 +114,6 @@ informative:
         ins: A. Perrig
         name: Adrian Perrig
         org: ETH Zuerich
-  I-D.path-properties-voc:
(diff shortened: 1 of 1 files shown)


nicorusti (approved):
Hi JC, IN see that you left several should etc in lowercase. Was this intentional, based on [rfc8174](https://datatracker.ietf.org/doc/html/rfc8174)?

I agree with you that some more work on these terms might be needed, I openet #12  to track this, as I'll merge this one to submit a new version to datatracker

jiceatscion:
As far as I remember, I went through every instance of must may should and their negatives. The should that I left in lowercase were likely intentional: layman use of the word for phrasing as opposed to normative, or sufficiently optional to not deserve an uppercase "should". But my opinion is just that, of course.

PR #13, opened 2024-02-27
Security considerations - DoS

nicorusti:


draft-dekater-scion-controlplane.md
```diff
@@ -1550,15 +1550,15 @@ To defend against this attack, methods to detect the wormhole attack are needed.
...
+Other mechanisms are required to avoid transport protocol attacks, where the attacker tries to exhaust the resources on a target server, such as a control service server, by opening many connections to this server. Possible means to mitigate this kind of DoS attacks are basically the same as for the current Internet, e.g., filtering, geo-blocking or using cookies. Thanks to its path-awareness, SCION enables more fine-grained filtering mechanisms, that can be employed to prevent misuse of critical request to the control plane:
 
 - ISD / AS whitelisting only allows requests from certain categories to pass the filter (e.g. requests from infrastructure nodes of the same AS, ISD, or from neighboring ASes)
 - Path length filtering discriminates packets based on the length of the traversed path (e.g. empty path for AS internal traffic, one hop paths limiting traffic to one-hop neighbors; and based on the number of path segments  to distinguish requests from external ISDs)
-- Rate limiting, that can be applied based on various identifiers (endpoint address, AS, or ISD)
+- Rate limiting, that can be applied on control plane RPCs that require authentication
```
matzf, line 1559: Suggestion:

---

[...] e.g. filtering, geo-blocking or using cookies.

RPC methods that are available to endpoints in the local AS are strictly separate from methods available to from other ASes. Specifically, expensive recursive path segment and trust material lookups are thus shielded from abuse by unauthorized entities.
For RPC methods exposed to other ASes, the control service implementation minimizes its attack surface by rejecting illegitimate callers based on ISD/AS, path type and length and any other available data points as soon as possible, i.e. immediately after determining the request type. For example:
- `SegmentCreationService.Beacon` can only be called by direct neighbors and thus calls from peers with a path length greater than one can immediately be discarded. 
- `SegmentRegistrationService.SegmentsRegistration` can only be called from within the same ISD, thus the source address must match the local ISD and the number of path segments must be 1.

nicorusti: Awesome, that is exactly what I was looking for. I introduced your points. I just slightly changed the wording of the first part  a bit (e.g. on local AS), but the concept should be unchanged.

PR #42, opened 2024-06-21
Scalability of path discovery

matzf:
Draft, some parts need to be converted from bullet points to prose.


draft-dekater-scion-controlplane.md
```diff
@@ -1297,6 +1309,64 @@ In comparison to these time scales, clock offsets in the order of minutes are im
...
+We cannot assume that the selected PCBs are strictly shortest paths through the network, but it's reasonable to assume that they will not be very much longer than the shortest paths either. It seems safe to assume that the PCBs are not longer than 10 hops on average for realistic networks.
+<-- TODO find something about _average_ shortest path. Probably can say something smaller than 10 -->
+
+With N the number of participating core ASes, an AS receives up to 5 * N PCBs per propagation interval per core link interface.
+For highly connected ASes, the number of PCBs received thus becomes rather large. In a network of 1000 ASes, an AS with 200 core links receives up to one million PCBs per propagation interval.
+This corresponds to roughly 150 thousand signature validations per second. This is just about the achievable throughput on a single core of a present day small server or desktop machine.
```
jiceatscion, line 1361: Isn't that supposed to be 1000000/5 = 200K?... well, that's assuming a 5 seconds propagation interval, but you said it was slower than that. So, I don't know. 150K amounts to a propagation interval of 6 and 2/3 seconds...I don't believe that you'd use that deliberately :-)

draft-dekater-scion-controlplane.md
```diff
@@ -1297,6 +1309,64 @@ In comparison to these time scales, clock offsets in the order of minutes are im
...
+In terms of bandwidth, this corresponds to very roughly 40MB/s.
```
jiceatscion, line 1362: Given your previous estimate of 250bytes per AS entry and core segments of ~10 ASes, I compute:
200000 * 10 * 250 = 500M/s. Even using your 150K/s figure, I still get 375M/s. I missed something?

draft-dekater-scion-controlplane.md
```diff
@@ -1297,6 +1309,64 @@ In comparison to these time scales, clock offsets in the order of minutes are im
...
+
+
+On a cold start of the network, path segments to each AS are discovered after a number of propagation steps proportional to the longest path. As mentioned, the longest path is typically not long. With a 5 second propagation period and a generous longest path of length 10, all path segments are discovered after 25 seconds on average.
+
+When a new parent-child link is added to the network, the parent AS will propagate the available PCBs in the next propagation interval. If the AS on the child side of the new link is a leaf AS, path discovery is thus complete after one single propagation interval. Otherwise, child ASes at distance D below the new link, learn of the new link after D further propagation steps.
+
+### Inter-ISD Beaconing
+In the inter-ISD core beaconing, PCBs are propagated omnidirectionally along core links. Each AS discovers path segments from itself to any other core AS.
+The number of distinct paths through the core network is typically very large. To keep the overhead managable, at most 5 path segments to every destination AS are discovered, and the propagation frequency is slower than in the intra-ISD beaconing.
```
jiceatscion, line 1353: Remind the reader what the typical interval would be?

draft-dekater-scion-controlplane.md
```diff
@@ -1297,6 +1309,64 @@ In comparison to these time scales, clock offsets in the order of minutes are im
...
+All of these depend on the the number and length of the discovered path segments, that is, on the total number of AS entries of the discovered path segments.
+
+Interesting metrics for scalability and speed of path discovery are the time until all discoverable path segments have been discovered after a "cold boot" and the time until a new link is usable.
+<!-- TODO: Note: removing links is not the job of the path discovery. For scheduled removal, path with old links expire. For link failures, paths are not renewed, routing around failure by path choice (data plane) -->
+Generally, the time until a specific PCB is built depends on its length and the propagation interval.
+At each AS, the PCB will be processed and propagated at the subsequent propagation interval.
+With a propagation interval T, the mean time until the PCB is propgated in one AS is T / 2 and the mean total time for the propagation steps of a PCB of length L is L * T / 2 (with a variance of L * T^2 / 12).
```
jiceatscion, line 1325: Here I'd cite the Irwin-Hall distribution, may be?

draft-dekater-scion-controlplane.md
```diff
@@ -1297,6 +1309,64 @@ In comparison to these time scales, clock offsets in the order of minutes are im
...
+<-- TODO find something about _average_ shortest path. Probably can say something smaller than 10 -->
+
+With N the number of participating core ASes, an AS receives up to 5 * N PCBs per propagation interval per core link interface.
+For highly connected ASes, the number of PCBs received thus becomes rather large. In a network of 1000 ASes, an AS with 200 core links receives up to one million PCBs per propagation interval.
+This corresponds to roughly 150 thousand signature validations per second. This is just about the achievable throughput on a single core of a present day small server or desktop machine.
+In terms of bandwidth, this corresponds to very roughly 40MB/s.
+
+
+On a cold start of the network, full connectivity is obtained after a number of propagation steps corresponding to the diameter of the network. With a generous diameter of 10, this corresponds to roughtly 5 minutes on average.
```
jiceatscion, line 1365: Ok, so here you seem to be using a 30s mean delay, so an interval of 1m.

draft-dekater-scion-controlplane.md
matzf: My calculation was:
- number of PCBs received per propagation interval: 5 * N * num interfaces = 1'000'000
- number of AS entries received (== number of signature validations) per propagation interval == ~10 * number of PCBs == 10'000'000
- number of AS entries per second: number of AS entries per propagation interval / 60s == 10'000'000 / 60 == 166'666.

I must have rounded pretty badly...

draft-dekater-scion-controlplane.md
matzf: :+1:

draft-dekater-scion-controlplane.md
jiceatscion: Ack. I had missed something.

draft-dekater-scion-controlplane.md
jiceatscion: Ok. Works for me. I had missed the multiplication by path length and I had the wrong interval.

draft-dekater-scion-controlplane.md
matzf: I've changed the numbers for the example; I noticed that length 10 is much longer than what we say above with the log n/loglog n average path length.
For 1000 ASes, the log n/loglog n is around 3.5. Because we say that we have paths that are longer than the shortest path, I use a generous 6 -- because this gives nice numbers.
Because the numbers coincidentally ended up confusingly similar to the intra-ISD case I've increased the number of core links from 200 to 300.

So:
5 * 300 * 1000 = 1.5 million PCBs
5 * 300 * 1000 * 6 / 60 = 150k AS entries per second
5 * 300 * 1000 * 6 * 250 / 60 = 37.5MB/s

draft-dekater-scion-controlplane.md
matzf: It seems Irwin-Hall was the wrong pointer after all.
Generally, the variance of the sum of independent random variables is just the sum of the variances;  the variance of the sum of n independent, _identically_ distributed variables is just n times the variance of a single variable.
The variance of continuously uniformly distributed random variables over interval [a,b] is (b-a)^/12. (!) -- in our case, this interval is [0,T] and so the variance for the individual random variables is T^2/12.

Even if this was not something I knew directly, this seems too basic / too much textbook knowledge to add a citation or long explanation. What do you think.

jiceatscion: .. which means that we don't actual care about T^2 / 12, do we? Anyway, you're right. The stats heads will know and the others won't care or understand (you just demonstrated this on me: I didn't understand).

jiceatscion (approved):

draft-dekater-scion-controlplane.md
```diff
@@ -1297,6 +1309,64 @@ In comparison to these time scales, clock offsets in the order of minutes are im
...
+To illustrate this with some numbers, an AS with a rather large number of 100 parent links receives at most 5000 PCBs during a propagation interval. Assuming a generous average length of 10 AS entries for these PCBs, this corresponds to 50000 AS entries. Due to the variable length fields in AS entries, the sizes for storage and transmission cannot be predicted exactly, and we'll assume an average of 250 bytes per AS entry. At the shortest, and thus chattiest, allowed propagation period of 5 seconds, this corresponds to a total bandwidth of very roughtly 2.5MB/s, and, processing 10000 signature verifications per second.
+If the same AS has 1000 child links, the propagation of the beacons will require signing one new AS entry for each of the propagated PCBs for each link, that is at most 50000 signatures per propagation interval.
```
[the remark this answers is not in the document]
matzf: Indeed, done. :+1:

matzf:
Thanks for fixing my typos, @nicorusti.

draft-dekater-scion-controlplane.md
```diff
@@ -1297,6 +1309,58 @@ In comparison to these time scales, clock offsets in the order of minutes are im
...
+
+To illustrate this with some numbers, an AS with a rather large number of 100 parent links receives at most 5000 PCBs during a propagation interval. Assuming a generous average length of 10 AS entries for these PCBs, this corresponds to 50000 AS entries. Due to the variable length fields in AS entries, the sizes for storage and transmission cannot be predicted exactly, and we'll assume an average of 250 bytes per AS entry. At the shortest, and thus chattiest, allowed propagation period of 5 seconds, this corresponds to a total bandwidth of very roughly 2.5MB/s, and, processing 10000 signature verifications per second.
+If the same AS has 1000 child links, the propagation of the beacons will require signing one new AS entry for each of the propagated PCBs for each link (at most 50 per link), that is at most 50000 signatures per propagation event.
+The total bandwidth for the propagation of these PCBs for all 1000 child links would, again very roughly, be around 25MB/s.
+All of these are manageable with even modest consumer hardware.
```
nicorusti, line 1340: Maybe we could mention that this is assuming a worst case scenario where  the AS is not doing any selection on received PCBs?

draft-dekater-scion-controlplane.md
```diff
@@ -1297,6 +1309,58 @@ In comparison to these time scales, clock offsets in the order of minutes are im
...
+
+On a cold start of the network, path segments to each AS are discovered after a number of propagation steps proportional to the longest path. As mentioned, the longest path is typically not long. With a 5 second propagation period and a generous longest path of length 10, all path segments are discovered after 25 seconds on average.
```
nicorusti, line 1342: In the draft we say:

> **Note:** Note that during bootstrapping and if the AS obtains a PCB containing a previously unknown path, the AS SHOULD forward the PCB immediately, to ensure quick connectivity establishment. 

Wouldn't this speed up bootstrapping time?

draft-dekater-scion-controlplane.md
```diff
@@ -1297,6 +1309,58 @@ In comparison to these time scales, clock offsets in the order of minutes are im
...
+
+When a new parent-child link is added to the network, the parent AS will propagate the available PCBs in the next propagation event. If the AS on the child side of the new link is a leaf AS, path discovery is thus complete after one single propagation interval. Otherwise, child ASes at distance D below the new link, learn of the new link after D further propagation steps.
```
nicorusti, line 1344:
```suggestion
When a new parent-child link is added to the network, the parent AS will propagate the available PCBs in the next propagation event. If the AS on the child side of the new link is a leaf AS, path discovery is thus complete after one single propagation interval. Otherwise, child ASes at distance D below the new link, learn of the new link after D further propagation intervals.
```

draft-dekater-scion-controlplane.md
```diff
@@ -1297,6 +1309,58 @@ In comparison to these time scales, clock offsets in the order of minutes are im
...
+We cannot assume that the selected PCBs are strictly shortest paths through the network, but it's reasonable to assume that they will not be very much longer than the shortest paths either.
+
+With N the number of participating core ASes, an AS receives up to 5 * N PCBs per propagation interval per core link interface.
+For highly connected ASes, the number of PCBs received thus becomes rather large. In a network of 1000 ASes, a highly connected AS with 300 core links receives up to 1.5 million PCBs per propagation interval.
+Assuming an average PCB length of 6 and the shortest propagation interval of 60 seconds, this corresponds to roughly 150 thousand signature validations per second. This throughput can be achieved on a single core of a present day small server or desktop machine.
+In terms of bandwidth, this corresponds to very roughly 38MB/s.
```
nicorusti, line 1356: I feel these calculations are a bit hypothetical, because: 
* This 38 MB/s of  bandwidth would be distributed across 300 links (maybe it is better to say that the bandwidth overhead is ~ 1 Mbit/s per link
* The overall CP load in an AS of such scale would be distributed on various instances (even tough one might suffice to handle the load for signature validation)

nicorusti (changes requested):
Thanks a lot for putting this together! Most of my comments are small language stuff, besides that I feel that some of the rough calculations could be framed slightly different

draft-dekater-scion-controlplane.md
nicorusti, line 1356: Maybe here we could summarize by saying that the overall message complexity for an AS is linear to the number of core ASes N.

draft-dekater-scion-controlplane.md
```diff
@@ -1135,11 +1135,11 @@ PCBs are propagated in batches to each connected downstream AS at a fixed freque
...
+Depending on the selection criteria, it may be necessary to keep more candidate PCBs than the *best PCBs set size* in the beacon store, to be able to determine the best set of PCBs. If this is the case, an AS should have a suitable pre-selection of candidate PCBs in place, in order to keep the beacon store capacity limited.
 
 - The *propagation interval* SHOULD be at least "5" (seconds) for intra-ISD beaconing and at least "60" (seconds) for core beaconing.
 
-**Note:** Note that during bootstrapping and if the AS obtains a PCB containing a previously unknown path, the AS SHOULD forward the PCB immediately, to ensure quick connectivity establishment.
+The scalability implications of such parameters are further discussed in [](#scalability). Note that during bootstrapping and if the AS obtains a PCB containing a previously unknown path, the AS SHOULD forward the PCB immediately, to ensure quick connectivity establishment.
```
jiceatscion, line 1142: New line missing before "Note that..." ?

jiceatscion (approved):

draft-dekater-scion-controlplane.md
```diff
@@ -1297,6 +1309,58 @@ In comparison to these time scales, clock offsets in the order of minutes are im
...
+All of these depend on the the number and length of the discovered path segments, that is, on the total number of AS entries of the discovered path segments.
+
+Interesting metrics for scalability and speed of path discovery are the time until all discoverable path segments have been discovered after a "cold boot", and the time until new link is usable.
+Generally, the time until a specific PCB is built depends on its length and the propagation interval.
+At each AS, the PCB will be processed and propagated at the subsequent propagation event. As propagation events are not synchronized between different ASes, a PCB arrives at a random point in time during the interval and is buffered before potentially being propagated.
+With a propagation interval T, the mean time until the PCB is propagated in one AS therefore is T / 2 and the mean total time for the propagation steps of a PCB of length L is L * T / 2 (with a variance of L * T^2 / 12).
+
+Note that link removal is not part of path discovery in SCION. For scheduled removal of links, operators let path segments expire. On link failures, end points route around the failed link by switching to different paths in the data plane.
```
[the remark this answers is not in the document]
matzf: I disagree, this seems entirely out of place. We shouldn't randomly throw in normative behavior rules into informative, explanatory sections.

draft-dekater-scion-controlplane.md
```diff
@@ -1297,6 +1309,58 @@ In comparison to these time scales, clock offsets in the order of minutes are im
...
+
+### Intra-ISD Beaconing
+In the intra-ISD beaconing, PCBs are propagated top-down, along parent-child links, from core to leaf ASes. Each AS discovers path segments from itself to the core ASes of its ISD.
+
+Typically, this directed, acyclic graph is narrow at the top, widens towards the leafs, and is relatively shallow; intermediate provider ASes have a large number of children, while they only have a small number of parents. The chain of intermediate providers from a leaf AS to a core AS is typically not long (e.g. local, regional, national provider, then core).
+
+Each AS potentially receives PCBs for all down paths between core to itself. While the number of distinct provider chains to the core is typically moderate, the multiplicity of links between provider ASes has multiplicative effect on the number of PCBs. Once this number grows above the limit value of 50, ASes trim the set of PCBs propagated. As the choice is among different ways to transit the local AS, operators are well equipped to choose among this set of PCBs.
```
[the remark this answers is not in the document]
matzf: Fixed, I made it "down-path segment" (and fixed the slightly broken sentence around it). FWIW, I think "down path" would not be hyphenated.

draft-dekater-scion-controlplane.md
```diff
@@ -1297,6 +1309,58 @@ In comparison to these time scales, clock offsets in the order of minutes are im
...
+Ultimately, the number of PCBs received by an AS per propagation interval remains bounded by 50 for each parent link of an AS, and at most 50 PCBs per child link are propagated. The length of these PCBs, and thus the number of AS entries to be processed and stored, is expected to be moderate and not grow considerably with network size. The total resource overhead for beacon propagation is easily manageable even for highly connected ASes.
```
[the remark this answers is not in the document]
matzf: Accepting to keep things consistent, but I find this SHOULD there rather strange. That's the one parameter that must be controlled in order that receivers of beacons won't be overwhelmed. Maybe it could say something like "MUST be less than 50 unless there is an out of band agreement". Anyway, maybe something for another day.

draft-dekater-scion-controlplane.md
```diff
@@ -1297,6 +1309,58 @@ In comparison to these time scales, clock offsets in the order of minutes are im
...
+
+To illustrate this with some numbers, an AS with a rather large number of 100 parent links receives at most 5000 PCBs during a propagation interval. Assuming a generous average length of 10 AS entries for these PCBs, this corresponds to 50000 AS entries. Due to the variable length fields in AS entries, the sizes for storage and transmission cannot be predicted exactly, and we'll assume an average of 250 bytes per AS entry. At the shortest, and thus chattiest, allowed propagation period of 5 seconds, this corresponds to a total bandwidth of very roughly 2.5MB/s, and, processing 10000 signature verifications per second.
```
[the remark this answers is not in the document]
matzf: > At one RTGWG session we got asked how SCION scales with the number of inter-AS links. It was someone working on Azure's backbone, they have ~30k + inter-AS links, so I feel 100 is not a rather large number

The 100 explicitly refer to links to (transit-)providers. I can't imagine that the Azure backbone has too many inter-AS links to transit providers.
Feel free to suggest a different way to present this, but what I tried was to give an impression for the upper bound for the resource overhead seen by a typical AS; 100 provider links seems more than what a typical (SCION-)AS will have.

draft-dekater-scion-controlplane.md
matzf: It's not making this assumption; in this scenario we received 50 * 100 == 5000 PCBs and forward only 50 of on each child link.

draft-dekater-scion-controlplane.md
matzf: That's the weirdest note. How can we define required / recommended behavior in a note?
Also, this behavior doesn't exist in the implementation. The idea seems a bit naive and completely underspecified. It's at least not obvious that this does not break the normal beaconing protocol, which _relies_ on these intervals.
I'd rather we just drop this note.

draft-dekater-scion-controlplane.md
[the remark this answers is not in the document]
matzf: I've not committed your suggestion directly but rather introduced the number one line above.
```diff
-Once this number grows above the limit value of 50, ASes trim the set of PCBs propagated.
+Once this number grows above the maximum recommended best PCBs set size of 50, ASes trim the set of PCBs propagated.
```

draft-dekater-scion-controlplane.md
matzf: But it's not, it's N times the path length. That's the whole buildup of this section:

- [Resource costs] depend on the the number and length of the discovered path segments, that is, on the total number of AS entries of the discovered path segments.

- Then we say that in core network, PCBs are roughly log(N) long.

- With N the number of participating core ASes, an AS receives up to 5 * N PCBs per propagation interval per core link interface.

matzf: The full bandwidth will be seen by the control service, both facing the network and for storage. This seems more relevant than the bandwidth across the individual links.

> The overall CP load in an AS of such scale would be distributed on various instances
I don't feel that this invalidates this calculations. The point was to illustrate that this is managable with modest resources.

I'll throw in a sentence to say that control service tasks can be distributed to scale to even larger ASes (like this 30k inter-domain link backbone).

draft-dekater-scion-controlplane.md
matzf: Done.

tzaeschke:
General Note: 
- The notes contain several references to "immediate cold-start PCB forwarding" which in turn refers to the section #selection where it says
under "Storing and Selecting Candidate PCBs":
"Note that during bootstrapping and if the AS obtains a PCB containing a previously unknown path, the AS SHOULD forward the PCB immediately, [...]."
- The notes contain several references to "beacon origination interval", i.e. the intervall at which new beacons are created. I couldn't find any description
  of this interval, maybe I overlooked it? Is it the same as the propagation interval?


# Some points I found could be useful to add to the doc:
- I couldn't find a discussion of beacon origination interval (or the RegistrationInterval), see "DefaultOriginationInterval" (or the DefaultRegistrationInterval) in the code.
  Maybe I overlooked it? 
- Load balancing? E.g. everybody picks the shortest paths, all other paths remain unused...? How is this handled?



# Introduction
## Avoiding Circular Dependencies and Partitioning
- Does this section (title) make sense as it is? The two topics (circular dependencies and partitioning) appear unrelated. Partitioning is discussed in a separate subsection
- Also, this section claims to contain a list that explains how circular dependencies are avoided.
  However, I am not sure how anything in the list explains anything about circular dependencies?

### Partition and Healing
- `ASes could always switch to otherwise unused links`. 

  - What are `unused` links? `Unused` indicates that they are not in the "Best PCB" set. 
  If that is the case, then we cannot simply "switch" to use them, we first need to have them discovered by beacons. This takes time, see `propagation interval` (unless PCB are forwarded immediately) and beacon `origination interval`.
  - Also: Does "Healing" include "adding new links"? In this case we also need to wait for the `propagation interval` and the `origination interval`.
  As I understand, with propagation interval being set to e.g. 10Minutes, this adds up to 5hops*1/2*10min= 25minutes for 5 hops.

# Path Exploration or Beaconing {#beaconing}
## Introduction and Overview

- `Intra-ISD beaconing`: Isn't this incomplete? How are DOWN segments created? I think leaf ASes need to propagate paths back to the COREs? -> Link to #intra-reg ?
- `Inter-ISD beaconing`: Similarily, I think there is a step where CORE beacons share their path DB with other CORE ASes in the same ISD?

### Extending a PCB
- `selects the best combinations`: Maybe link to a section that explains how this works? And what "best" means?

## Path-Segment Construction Beacons
### PCB Validity
- `For the purpose of validation, a timestamp is considered "future" if it is later than the current time at the point of validation plus the minimum expiration time of a hop field (337.5 seconds, see ).`
Maybe add an explanation why we add the minimum expiration time here?
Shouldn't "future" simply be timestamp + some_delta_to_account_for_server_time_inaccuracies, where the delta is maybe a few seconds rather than 5.5 minutes?




## Propagation of PCBs {#path-prop}
### Selection of PCBs to Propagate {#selection}
#### Storing and Selecting Candidate PCBs

- `temporary storage`: Maybe clarify what "temporary" means. 
  - How long are PCBs stored? 
  - Under what circumstances are they removed, e.g. when they expire? 
  - Or possibly replaced with a new version? What if the new version has an earlier expiration date?


- `At each propagation event, each AS selects a set of the best PCBs from the candidates in the beacon store`


- `The best PCBs set size SHOULD be at most "50" (PCBs) for intra-ISD beaconing and at most "5" (PCBs) for core beaconing.`
  - I found this a bit confusing: the "50" appears to be the total number of beacons forwarded for non-core ASes, whereas the "5" refers to the number of 
    PCBs _per_ remote CORE AS. Maybe clarify this?

- `Note that during bootstrapping and if the AS obtains a PCB containing a previously unknown path, the AS SHOULD forward the PCB immediately, [...].`
  - Is this true? It appears to conflict with many other parts of the document that talk about propagation intervals in the contect of cold-start.
  - Is there a difference between "bootstrapping" (used here) and "cold-start" (used in other places)? Maybe stick to one term or explicitly declare equality?
  - Is this subject to the "Best 5 PCB" rule? -> If a new path is immediately forwarded, does it count towards th "best 5"? If not, then we are effectively
    forwarding >5 paths, correct? If yes, then the first 5 paths are always the best until they expire and can be replaced with other paths? 
  - What means "unknown path"? Does it refer to all links in the segment or just the remote AS? If it is all links in the segment, then there may be many new paths
    coming in all the time that need to be forwarded immediately, or not?
    


### Effects of Clock Inaccuracy
- `PCBs are propagated at a configurable interval (typically, one minute).`
  - Unless they are new, then the interval is ignored, see "immediate cold-start PCB forwarding".
  - Maybe rephrase: "(typically, one minute)" ---> "immediately for new beacons, minimum 5secs for intra-ISD, minimum one minute for inter-ISD"?
  See #path-prop: "The propagation interval SHOULD be at least "5" (seconds) for intra-ISD beaconing and at least "60" (seconds) for core beaconing.".
  - Also, AFAIK, the current configured real-world interval is more like 10-15 minutes...?  

- `PCBs with N hops may be validated up to N intervals (so typically N minutes) after origination`
  - I think the word "typically" is misleading here, it can be understood as "PCBs [...] are validated typically after N minutes", whereas it actually means that the _maximum_ is typically N minutes. 
    Rephrase to "(maximally N minutes)" or "(amounting to N minutes, assuming the minimum inter-ISD propagation interval)"

- Rephrase `The norm is 6 hours.` to `... SHOULD be 6 hours` ? What does 'norm' mean?

- `In comparison to these time scales, clock offsets in the order of minutes are immaterial.`  
  - This relates only to the previous paragraph about certificates; I guess it should be attached to the previous paragraph?


## Path Discovery Time and Scalability {#scalability}
- ~`balances _OF_ the number of discovered paths` -> remove OF~ Already fixed

- `Generally, the time until a specific PCB is built depends on its length and the propagation interval.`? 
   - I think in the context of "cold boot¨, the propagation delay is "0", see "immediate cold-start PCB forwarding".
- `PCB arrives at a random point in time during the interval and is buffered before potentially being propagated`
  - see "immediate cold-start PCB forwarding"
  - Also, I think the calculation needs to take into account the "beacon origination interval".

- `As will become apparent, the inter-ISD beaconing results in excessive overhead with very large numbers of participating core ASes.`
  - Does this need to be in the IETF spec?
- `The ideal topology for SCION is to keep the inter-ISD core network to a moderate size, to benefit from the divide-and-conquer partitioning of ASes into ISDs and the efficiency of the intra-ISD beaconing.`
  - What is done to ensure this? What happens if the size is not moderate? What is "moderate"?


### Intra-ISD Beaconing

- `Otherwise, child ASes at distance D below the new link, learn of the new link after D further propagation steps`
   - New path: "immediate cold-start PCB forwarding"


### Inter-ISD Beaconing

- `On a cold start of the network, [...]. With a 5 second propagation period [...]`
   - Above it says that bootstrapping results in immediatede forwarding, see "immediate cold-start PCB forwarding".

- `When a new link is added to the network, it will be available to connect two ASes at distances from the link D1 and D2 from the link, respectively, after a mean time (D1+D2)*T/2.`
  - Typo: duplicated "from the link";
  - Also: see previous point about "immediate cold-start PCB forwarding"

draft-dekater-scion-controlplane.md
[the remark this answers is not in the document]
nicorusti: My intention was to refer to highlight that draining links by letting segments expire is recommended, but I see how this is off topic here. No strong opinions here, I'll leave as it it.

draft-dekater-scion-controlplane.md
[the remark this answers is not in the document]
nicorusti: This is hyphenated in the rest of the draft, thank you for keeping it consistent!

draft-dekater-scion-controlplane.md
nicorusti: Right, thanks for clarifying!

draft-dekater-scion-controlplane.md
nicorusti: Removed. This comes from the footnote page 24 of the SCION book, it was edited by Markus/Sam. Any idea how this ended up there?

draft-dekater-scion-controlplane.md
nicorusti: Thanks for the clarification, as far as I understand then the message complexity in terms of number of signature validations per AS is O(N*log(N)), while the amount of propagated PCBs per AS is  O(N), correct?

nicorusti:
Thank you for your feedback @tzaeschke ! I respond here regarding the scalability and clock inaccuracy sections.  For other sections, and for points that we don't have time to address in time this revision, I opened separate issues: 
* https://github.com/scionassociation/scion-cp_I-D/issues/45 
* https://github.com/scionassociation/scion-deployment_I-D/issues/1  
* https://github.com/scionassociation/scion-cp_I-D/issues/46 


## Regarding Effects of Clock Inaccuracy

> I think the word "typically" is misleading here, it can be understood as "PCBs [...] are validated typically after N minutes", whereas it actually means that the maximum is typically N minutes.
Rephrase to "(maximally N minutes)"

Done, maximally N minutes sounds good.

> Also, AFAIK, the current configured real-world interval is more like 10-15 minutes...?
 
@matzf I is it 1 min as in the draft, or 10-15? 10-15 feels a bit high to me 


>  Rephrase `The norm is 6 hours.` to `... SHOULD be 6 hours` ? What does 'norm' mean? 

I am a bit reluctant to use RFC2119 language (uppercase SHOULD) for exactly 6 hours. This is a value that overall depends on the maximum AS path expected in the network, and it might as well be a different value. I therefore rephrased like this:
`For this reason, it is unadvisable to create hops with a short expiration time, that should be around 6 hours.` 

> In comparison to these time scales, clock offsets in the order of minutes are immaterial.
> This relates only to the previous paragraph about certificates; I guess it should be attached to the previous paragraph?

Done.



## Regarding Path Discovery Time and Scalability {#scalability} 

> The notes contain several references to "immediate cold-start PCB forwarding"  

@matzf clarified [here](https://github.com/scionassociation/scion-cp_I-D/pull/42#discussion_r1666689901) that this is not the case, I removed that note, this should also solve some many of the consistency issues you reported.

> Also, I think the calculation needs to take into account the "beacon origination interval".
 
 To be handled in https://github.com/scionassociation/scion-cp_I-D/issues/45
 

> `As will become apparent, the inter-ISD beaconing results in excessive overhead with very large numbers of participating core ASes.` 
Does this need to be in the IETF spec?

Good point, I rephrased this section to: 
```To achieve scalability in its routing process, SCION uses a divide-and-conquer approach, partitioning  ASes into ISDs. In order to benefit from this, an ideal topology SCION should keep the inter-ISD core network to a moderate size. For more specific observations, we distinguish between intra- and inter-ISD beaconing. ```

> What is done to ensure this? What happens if the size is not moderate? What is "moderate"?
 
 We give some numbers in the  Inter-ISD Beaconing section with an example with 1000 core ASes, this gives a rough figure. The bandwidth and computation overhead figures there should also give a rough hint of what happens if the network grows too much: the overhead becomes considerable. 
 What is done to ensure this IMHO depends on how the network is deployed, I think this topic would be a better fit to be discussed in the new deployment Internet Draft, I opened an issue there: https://github.com/scionassociation/scion-deployment_I-D/issues/1  
 
 > Typo: duplicated "from the link";

Fixed

nicorusti (approved):

matzf:
Commented on the related issues for the other sections.

> > Also, AFAIK, the current configured real-world interval is more like 10-15 minutes...?
> 
> @matzf I is it 1 min as in the draft, or 10-15? 10-15 feels a bit high to me

The 1 minute value seems realistic. SCIONLab uses 5 seconds for non-core beaconing and 1 minute for core beaconing. Anapaya's infrastructure reportedly runs with 30s.

PR #43, opened 2024-06-21
clarify path reversibility

### LINKED ISSUE #22: Clarify path reversal for symmetric routing
think in the provider world it is expected that traffic between two endpoints uses bidirectionally the same path symmetrically. As each AS can select its own path to other AS's, it looks like SCION is not symmetrical. Maybe a comment around that could help.
 
  @nicorusti: a sentence to clarify this could be added in 1.6. at bullet point path reversal.

nicorusti:
Resolves #22

### DIFF
--- a/draft-dekater-scion-controlplane.md
+++ b/draft-dekater-scion-controlplane.md
@@ -375,7 +375,7 @@ A secure and reliable routing architecture has to be designed specifically to av
 
 - Neighbor-based path discovery: Path discovery in SCION is performed by the beaconing mechanism. In order to participate in this process, an AS only needs to be aware of its direct neighbors. As long as no path segments are available, communicating with the neighboring ASes is possible with the one-hop path type, which does not rely on any path information. SCION uses these *one-hop paths* to propagate PCBs to neighboring ASes to which no forwarding path is available yet. The One-Hop Path Type is described in more detail in {{I-D.scion-dp}}.
 - Path segment types: SCION uses different types of path segments to compose end-to-end paths. Notably, a single path segment already enables intra-ISD communication. For example, a non-core AS can reach the core of the local ISD simply by using an up-segment fetched from the local path storage, which is populated during the beaconing process.
-- Path reversal: In SCION, every path is reversible—i.e., the receiver of a packet can reverse the path in the packet header to send back a reply packet without having to perform a path lookup and ensuring that the packet follows the same path symmetrically.
+- Path reversal: In SCION, every path is reversible. That is, the receiver of a packet can reverse the path in the packet header in order to produce a reply packet without having to perform a path lookup. Such a packet follows the original packet's path backwards.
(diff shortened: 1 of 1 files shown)


draft-dekater-scion-controlplane.md
```diff
@@ -375,7 +375,7 @@ A secure and reliable routing architecture has to be designed specifically to av
...
 - Path segment types: SCION uses different types of path segments to compose end-to-end paths. Notably, a single path segment already enables intra-ISD communication. For example, a non-core AS can reach the core of the local ISD simply by using an up-segment fetched from the local path storage, which is populated during the beaconing process.
-- Path reversal: In SCION, every path is reversible—i.e., the receiver of a packet can reverse the path in the packet header to send back a reply packet without having to perform a path lookup.
+- Path reversal: In SCION, every path is reversible—i.e., the receiver of a packet can reverse the path in the packet header to send back a reply packet without having to perform a path lookup and ensuring that the packet follows the same path symmetrically.
```
jiceatscion, line 378: That sentence doesn't seem grammatically correct. I am not sure what is doing the "ensuring". How about:"

- Path reversal: In SCION, every path is reversible. That is, the receiver of a packet can reverse the path in the packet header in order to produce a reply packet without having to perform a path lookup. Such a packet follows the original packet's path backwards.
"

draft-dekater-scion-controlplane.md
```diff
@@ -1423,15 +1423,16 @@ An endpoint (source) that wants to start communication with another endpoint (de
...
 3. Next, the control service of the source AS combines up-path segments with the newly retrieved core-path segments. The control service then queries the control services of the remote core ASes in the destination ISD, to fetch down-path segments to the destination AS. To reach the remote core ASes, the control service of the source AS uses the previously obtained and combined up- and core segments.
 4. Finally, the control service of the source AS returns all retrieved path segments to the source endpoint.
-5. Once it has obtained all path segments, the endpoint combines them into an end-to-end path in the data plane.
+5. Once it has obtained all path segments, the source endpoint combines them into an end-to-end path in the data plane.
+6. The destination endpoint, once it receives the first packet, may revert the path in the received packet to respond without needing to perform a further path lookup. This also ensures that traffic flows on the same path bidirectionally.
```
jiceatscion, line 1435: Point 6: haven't we already explained that? May be we could be more laconic here:

6. The destination endpoint, once it receives the first packet, has the option to revert the path in the received packet in order to construct a response.

draft-dekater-scion-controlplane.md
nicorusti: Thanks a lot, nice wording!

draft-dekater-scion-controlplane.md
nicorusti: I was a b it redundant with wording because of the feedback we received: 

> think in the provider world it is expected that traffic between two endpoints uses bidirectionally the same path symmetrically. As each AS can select its own path to other AS's, it looks like SCION is not symmetrical. Maybe a comment around that could help. 

I'll use your proposal for more succinct wording

jiceatscion (approved):

Issue #12, opened 2024-02-23
Continue improving use of MUST and SHOULD according to rfc2119 and rfc8174


nicorusti:
Continue work done in #11: 
       - MUST: Without this, the protocol will not work
       - SHOULD: Should do this unless you have a very good reason not to
       - MAY: Try to not use "MAY", if not needed
      

etc.

nicorusti:
Let's fix the: "Allowed", review use of MAY, and then resolve this

nicorusti:
While we added words,  we need to look at capitalization of these words, this is still not right in all places:

From rfc8174: 
> The words have the meanings specified herein only when they are in all capitals."

PR #44, opened 2024-07-04
Highlighted use of MUST, MUST NOT and SHOULD where appropriate.

knmeynell:
Have not changed this in code references
Solves #12

### DIFF
--- a/draft-dekater-scion-controlplane.md
+++ b/draft-dekater-scion-controlplane.md
@@ -1403,7 +1403,7 @@ The control service of a non-core AS has to register the newly created down-segm
    message SegmentsRegistrationResponse {}
 ~~~~
 
-- `SegmentType`: Specifies the type of the path segment that MUST be registered. Currently, only the following type is used:
+- `SegmentType`: Specifies the type of the path segment to be registered. Currently, only the following type is used:
   - `SEGMENT_TYPE_DOWN`: Specifies a down-segment.
 - `map<int32, Segments> segments`: Represents a separate list of segments for each path segment type. The key is the integer representation of the corresponding `SegmentType`.
 
--- a/draft-dekater-scion-controlplane.md
+++ b/draft-dekater-scion-controlplane.md
@@ -345,7 +345,7 @@ In SCION, the following rules apply:
 - The `::` zero-compression feature of IPv6 MUST NOT be used. The feature has very limited use in a 48-bit address space and would only add more complexity.
 - A range of AS numbers can be shortened with a notation similar to the one used for CIDR IP ranges ({{RFC4632}}). For example, the range of the lowest 32-bit AS numbers (0-4294967295) can be represented as `0:0:0/16`.
 
-For historical reasons, SCION AS numbers in the lower 32 bit range may also be represented as decimal for human readability. For example, if a program receives the AS number `0:1:f`, it may display the number as "65551".
+For historical reasons, SCION AS numbers in the lower 32 bit range MAY also be represented as decimal for human readability. For example, if a program receives the AS number `0:1:f`, it MAY display the number as "65551".
 
 
 #### Special-Purpose SCION AS Numbers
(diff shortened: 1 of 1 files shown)


draft-dekater-scion-controlplane.md
```diff
@@ -919,7 +919,7 @@ The signature Sig<sub>i</sub> of an AS entry ASE<sub>i</sub> is now computed as
 Sig<sub>i</sub> =
 K<sub>i</sub>( SegInfo || ASE<sub>0</sub><sup>(signed)</sup> || Sig<sub>0</sub> || ... || ASE<sub>i-1</sub><sup>(signed)</sup> || Sig<sub>i-1</sub> || ASE<sub>i</sub><sup>(signed)</sup> )
 
-The signature metadata minimally contains the ISD-AS number of the signing entity and the key identifier of the public key that should be used to verify the message. For more information on signing and verifying control-plane messages, see the chapter "Signing and Verifying Control-Plane Messages" of the SCION Control-Plane PKI Specification {{I-D.scion-cppki}}.
+The signature metadata minimally contains the ISD-AS number of the signing entity and the key identifier of the public key that SHOULD be used to verify the message. For more information on signing and verifying control-plane messages, see the chapter "Signing and Verifying Control-Plane Messages" of the SCION Control-Plane PKI Specification {{I-D.scion-cppki}}.
```
nicorusti, line 922: I think this should be a MUST, however this should be regulated in the PKI dtaft. I''ll just reword to avoid using RFC2119 words
```suggestion
The signature metadata minimally contains the ISD-AS number of the signing entity and the key identifier of the public key to be used to verify the message. For more information on signing and verifying control-plane messages, see the chapter "Signing and Verifying Control-Plane Messages" of the SCION Control-Plane PKI Specification {{I-D.scion-cppki}}.
```

draft-dekater-scion-controlplane.md
```diff
@@ -1230,7 +1230,7 @@ The propagation process in intra-ISD beaconing includes the following steps:
    - The ingress interface to this AS, in the hop field component.
    - The egress interface to the neighboring AS, also in the hop field component.
    - The ISD_AS number of the next AS, in the signed body component of the AS entry.
-   - If the AS has peering links, the control service should add corresponding peer entry components to the signed body of the AS entry; one peer entry component for each peering link that the AS wants to advertise. The hop field component of each added peer entry MUST have a specified egress interface.
+   - If the AS has peering links, the control service SHOULD add corresponding peer entry components to the signed body of the AS entry; one peer entry component for each peering link that the AS wants to advertise. The hop field component of each added peer entry MUST have a specified egress interface.
```
nicorusti, line 1233: I think this is a MUST rather than a SHOULD

draft-dekater-scion-controlplane.md
```diff
@@ -1318,14 +1318,14 @@ The up- and down-segments do not have to be equal. An AS may want to communicate
...
 
 1. The control service adds a new AS entry to the PCB. This new AS entry MUST be defined as follows:
    - The next AS MUST NOT be specified.
      - In Protobuf message format, this means that the value of the `next_isd_as` field in the `ASEntrySignedBody` component MUST be "0".
    - The egress interface in the hop field component MUST NOT be specified.
      - In Protobuf message format, this means that the value of the `egress` field in the `HopField` component MUST be "0".
-2. If the AS has peering links, the control service should add corresponding peer entry components to the signed body of the AS entry - one peer entry component for each peering link that the AS wants to advertise. The egress interface ID in the hop field component of each added peer entry MUST NOT be specified.
+2. If the AS has peering links, the control service SHOULD add corresponding peer entry components to the signed body of the AS entry - one peer entry component for each peering link that the AS wants to advertise. The egress interface ID in the hop field component of each added peer entry MUST NOT be specified.
```
nicorusti, line 1328: Same as above

draft-dekater-scion-controlplane.md
```diff
@@ -1403,7 +1403,7 @@ The control service of a non-core AS has to register the newly created down-segm
    message SegmentsRegistrationResponse {}
 ~~~~
 
-- `SegmentType`: Specifies the type of the path segment that must be registered. Currently, only the following type is used:
+- `SegmentType`: Specifies the type of the path segment that MUST be registered. Currently, only the following type is used:
```
nicorusti, line 1406: I feel that registering the segment is not an absolute requirement to the spec, I'd avoid RFC2119 wording: 
```suggestion
- `SegmentType`: Specifies the type of the path segment to be registered. Currently, only the following type is used:
```

draft-dekater-scion-controlplane.md
nicorusti: Discussed with @matzf , MAY seems more appropriate, as this is somehow optional if the AS wants to advertise the peering link

```suggestion
   - If the AS has peering links, the control service MAY add corresponding peer entry components to the signed body of the AS entry; one peer entry component for each peering link that the AS wants to advertise. The hop field component of each added peer entry MUST have a specified egress interface.
```

draft-dekater-scion-controlplane.md
nicorusti:
```suggestion
2. If the AS has peering links, the control service MAY add corresponding peer entry components to the signed body of the AS entry - one peer entry component for each peering link that the AS wants to advertise. The egress interface ID in the hop field component of each added peer entry MUST NOT be specified.
```

nicorusti (approved):

Issue #48, opened 2024-07-08
Clarify beaconing fast retry at bootstrapping


nicorusti:
In #42 we discussed about  propagation interval in case of "bootstrapping"/"cold-start". Following clarifications in this [slack discussion](https://scionproto.slack.com/archives/C8ADA9CEP/p1720431398661119), it turns out that: 
> There is a fast recovery where we attempt to beacon a lot more frequently until we have beaconed successfully over the given interface

This is implemented in [here](https://github.com/scionproto/scion/blob/a1ed6246ecd6beae9c82544032de4f9fecab1058/control/beaconing/propagator.go#L324) in scionproto. There was just one note in the draft that got removed in #42. We should re-add this, together with some clarifications, especially in the _Effects of Clock Inaccuracy_ and _Path Discovery Time and Scalability_ sections.

Some points to be addressed were reported by @tzaeschke [here](https://github.com/scionassociation/scion-cp_I-D/pull/42#issuecomment-2211084556): 

> ### Storing and Selecting Candidate PCBs
>- `Note that during bootstrapping and if the AS obtains a PCB containing a previously unknown path, the AS SHOULD forward the PCB immediately, [...].`
>   - Is this true? It appears to conflict with many other parts of the document that talk about propagation intervals in the contect of cold-start.
>   - Is there a difference between "bootstrapping" (used here) and "cold-start" (used in other places)? Maybe stick to one term or explicitly declare equality?
>   - Is this subject to the "Best 5 PCB" rule? -> If a new path is immediately forwarded, does it count towards th "best 5"? If not, then we are effectively
>     forwarding >5 paths, correct? If yes, then the first 5 paths are always the best until they expire and can be replaced with other paths? 
>   - What means "unknown path"? Does it refer to all links in the segment or just the remote AS? If it is all links in the segment, then there may be many new paths
    coming in all the time that need to be forwarded immediately, or not?
    
> ### Effects of Clock Inaccuracy
> - `PCBs are propagated at a configurable interval (typically, one minute).`
>   - Unless they are new, then the interval is ignored, see "immediate cold-start PCB forwarding".
>   - Maybe rephrase: "(typically, one minute)" ---> "immediately for new beacons, minimum 5secs for intra-ISD, minimum one minute for inter-ISD"?

> ### Path Discovery Time and Scalability {#scalability}
> - `Generally, the time until a specific PCB is built depends on its length and the propagation interval.`? 
[...]

PR #49, opened 2024-07-08
Clarify beaconing fast retry at bootstrapping

nicorusti:
Solves #48

### DIFF
--- a/draft-dekater-scion-controlplane.md
+++ b/draft-dekater-scion-controlplane.md
@@ -1139,7 +1139,12 @@ Depending on the selection criteria, it may be necessary to keep more candidate 
 
 - The *propagation interval* SHOULD be at least "5" (seconds) for intra-ISD beaconing and at least "60" (seconds) for core beaconing.
 
-Note that during bootstrapping and if the AS obtains a PCB containing a previously unknown path, the AS SHOULD forward the PCB more frequently, to ensure quick connectivity establishment.
+Note that under certain conditions, to ensure quick connectivity establishment, an AS MAY attempt to forward a PCB more frequently:
+
+- if no beacon was successfully sent over the interface in the current propagation interval (e.g. because the corresponding RPC failed)
+
+- or if the AS receives a previously unknown path segment
+
 The scalability implications of such parameters are further discussed in [](#scalability).
 
 
@@ -1302,7 +1307,7 @@ A PCB originated by a given control service is validated by all the control serv
 * A fast clock at origination or a slow clock at reception will yield a lengthened expiration time for hops, and possibly an origination time in the future.
 * A slow clock at origination or a fast clock at reception will yield a shortened expiration time for hops, and possibly an expiration time in the past.
 
-This bias comes in addition to a structural delay: PCBs are propagated at a configurable interval (typically, around one minute), except at AS bootstrapping. As a result of this and the way they are iteratively constructed, PCBs with N hops may be validated at worst case up to N intervals (so maximally N minutes) after origination. This creates a constraint on the expiration of hops. Hops of the minimal expiration time (337.5 seconds - see [](#hopfield)) would render useless any PCB describing a path longer than 5 hops. For this reason, it is unadvisable to create hops with a short expiration time, that should be around 6 hours.
+This bias comes in addition to a structural delay: PCBs are propagated at a configurable interval (typically, around one minute). As a result of this and the way they are iteratively constructed, PCBs with N hops may be validated at worst case up to N intervals (so maximally N minutes) after origination. This creates a constraint on the expiration of hops. Hops of the minimal expiration time (337.5 seconds - see [](#hopfield)) would render useless any PCB describing a path longer than 5 hops. For this reason, it is unadvisable to create hops with a short expiration time, that should be around 6 hours.
 
 The control service and its clients authenticate each-other according to their respective AS's certificate. Path segments are authenticated based on the certificates of the ASes that they refer to. The expiration of a SCION AS certificate typically ranges from 3h to 5 years.
 In comparison to these time scales, clock offsets in the order of minutes are immaterial.
@@ -1317,7 +1322,7 @@ The path discovery mechanism balances the number of discovered paths and the tim
 The resource costs for path discovery are communication overhead, processing and storage. Communication is transmitting the PCBs and occasionally obtaining the required PKI material. Processing cost is validating the signatures of the AS entries, signing new AS entries, and, to a lesser extent, evaluating the beaconing policies. Storage is both the temporary storage of PCBs before the next propagation interval, and the storage of complete discovered path segments.
 All of these depend on the the number and length of the discovered path segments, that is, on the total number of AS entries of the discovered path segments.
 
-Interesting metrics for scalability and speed of path discovery are the time until all discoverable path segments have been discovered after a "cold boot", and the time until new link is usable.
+Interesting metrics for scalability and speed of path discovery are the time until all discoverable path segments have been discovered after a "cold start", and the time until new link is usable.
 Generally, the time until a specific PCB is built depends on its length and the propagation interval.
 At each AS, the PCB will be processed and propagated at the subsequent propagation event. As propagation events are not synchronized between different ASes, a PCB arrives at a random point in time during the interval and is buffered before potentially being propagated.
 With a propagation interval T at each AS, the mean time until the PCB is propagated in one AS therefore is T / 2 and the mean total time for the propagation steps of a PCB of length L is L * T / 2 (with a variance of L * T^2 / 12).
@@ -1339,7 +1344,7 @@ If the same AS has 1000 child links, the propagation of the beacons will require
 The total bandwidth for the propagation of these PCBs for all 1000 child links would, again very roughly, be around 25MB/s.
 All of these are manageable with even modest consumer hardware.
 
-On a cold start of the network, path segments to each AS are discovered after a number of propagation steps proportional to the longest path. As mentioned, the longest path is typically not long. With a 5 second propagation period and a generous longest path of length 10, all path segments are discovered after 25 seconds on average.
+On a cold start of the network, path segments to each AS are discovered at worst after a number of propagation steps proportional to the longest path. As mentioned, the longest path is typically not long. With a 5 second propagation period and a generous longest path of length 10, all path segments are discovered after 25 seconds on average.
 
 When a new parent-child link is added to the network, the parent AS will propagate the available PCBs in the next propagation event. If the AS on the child side of the new link is a leaf AS, path discovery is thus complete after one single propagation interval. Otherwise, child ASes at distance D below the new link, learn of the new link after D further propagation intervals.
 
@@ -1357,7 +1362,7 @@ All of these are manageable on a present day small server or desktop machine.
 For much larger, more highly connected ASes, the path-discovery tasks of the control service can be distributed over many instances in order to increase the PCB throughput.
 
 
-On a cold start of the network, full connectivity is obtained at worst after a number of propagation steps corresponding to the diameter of the network, unless ASes propagate beacons more frequently at bootstrapping. Assuming a network diameter of 6, this corresponds to roughly 3 minutes on average.
+On a cold start of the network, full connectivity is obtained at worst after a number of propagation steps corresponding to the diameter of the network. Assuming a network diameter of 6, this corresponds to roughly 3 minutes on average.
 
 When a new link is added to the network, it will be available to connect two ASes at distances D1 and D2 from the link, respectively, after a mean time (D1+D2)*T/2.
 
--- a/draft-dekater-scion-controlplane.md
+++ b/draft-dekater-scion-controlplane.md
@@ -1139,6 +1139,7 @@ Depending on the selection criteria, it may be necessary to keep more candidate 
 
 - The *propagation interval* SHOULD be at least "5" (seconds) for intra-ISD beaconing and at least "60" (seconds) for core beaconing.
 
+Note that during bootstrapping and if the AS obtains a PCB containing a previously unknown path, the AS SHOULD forward the PCB more frequently, to ensure quick connectivity establishment.
 The scalability implications of such parameters are further discussed in [](#scalability).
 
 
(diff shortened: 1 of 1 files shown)


draft-dekater-scion-controlplane.md
```diff
@@ -1139,6 +1139,7 @@ Depending on the selection criteria, it may be necessary to keep more candidate
 
 - The *propagation interval* SHOULD be at least "5" (seconds) for intra-ISD beaconing and at least "60" (seconds) for core beaconing.
 
+Note that during bootstrapping and if the AS obtains a PCB containing a previously unknown path, the AS SHOULD forward the PCB more frequently, to ensure quick connectivity establishment.
```
jiceatscion, line 1142: I side with Matthias regarding the phrase "bootstrapping". This calls for a definition and such is hard to come by. Do we need to mention that at all?

Regarding "The AS SHOULD". We just said that it "SHOULD not" in the paragraph above. So can't possibly express it like that. This may be one case where "MAY" is approriate, assuming we specify under which condition (since that derogates from the previous paragraph.

Regarding when it MAY... I think that "previously unknown" segments, or "previously unreachable destinations" could qualify.

draft-dekater-scion-controlplane.md
```diff
@@ -1301,7 +1302,7 @@ A PCB originated by a given control service is validated by all the control serv
...
+This bias comes in addition to a structural delay: PCBs are propagated at a configurable interval (typically, around one minute), except at AS bootstrapping. As a result of this and the way they are iteratively constructed, PCBs with N hops may be validated at worst case up to N intervals (so maximally N minutes) after origination. This creates a constraint on the expiration of hops. Hops of the minimal expiration time (337.5 seconds - see [](#hopfield)) would render useless any PCB describing a path longer than 5 hops. For this reason, it is unadvisable to create hops with a short expiration time, that should be around 6 hours.
```
jiceatscion, line 1305: Same consideration about boostrapping. What is bootstrapping.

draft-dekater-scion-controlplane.md
```diff
@@ -1356,7 +1357,7 @@ All of these are manageable on a present day small server or desktop machine.
 For much larger, more highly connected ASes, the path-discovery tasks of the control service can be distributed over many instances in order to increase the PCB throughput.
 
 
-On a cold start of the network, full connectivity is obtained after a number of propagation steps corresponding to the diameter of the network. Assuming a network diameter of 6, this corresponds to roughly 3 minutes on average.
+On a cold start of the network, full connectivity is obtained at worst after a number of propagation steps corresponding to the diameter of the network, unless ASes propagate beacons more frequently at bootstrapping. Assuming a network diameter of 6, this corresponds to roughly 3 minutes on average.
```
jiceatscion, line 1360: And again. Here I think it would be enough to mention the worst case and not bother with "unless...".

draft-dekater-scion-controlplane.md
nicorusti: "Bootstrapping" was a quick & dirty word, I'll replace it with a list of conditions when the propagation interval MAY not be respected.

draft-dekater-scion-controlplane.md
nicorusti: Ok, I will remove all mentions of bootstrapping. However, I'd like that we discuss (maybe also with oncilla) if this fast retry mechanism should really be optional, or perhaps mandatory. 
If it stays optional (using MAY in comment above), then we are ok to say here `at worst case`. If it becomes mandatory, then the structural delay here might become a lot better (however, I feel we still don't understand the full implications).

draft-dekater-scion-controlplane.md
```diff
@@ -1139,6 +1139,12 @@ Depending on the selection criteria, it may be necessary to keep more candidate
 
 - The *propagation interval* SHOULD be at least "5" (seconds) for intra-ISD beaconing and at least "60" (seconds) for core beaconing.
 
+Note that under certain conditions, to ensure quick connectivity establishment, an AS MAY attempt to forward a PCB more frequently:
+
+- if no beacon was successfully sent over the interface in the current propagation interval (e.g. because the corresponding RPC failed)
+
+- or if the AS receives a previously unknown path segment
+
```
nicorusti, line 1147: @jiceatscion @matzf @oncilla do these conditions sound reasonable? How would the AS check that it receives a previously unknown segment, shall we be more precise about this? 

Also, from Tillman's comments: 
>  Is this subject to the "Best 5 PCB" rule? -> If a new path is immediately forwarded, does it count towards th "best 5"? If not, then we are effectively

draft-dekater-scion-controlplane.md
nicorusti: (fee free to resolve if you like my text, or to propose an alternative)

draft-dekater-scion-controlplane.md
nicorusti: (fee free to resolve if you like my text, or to propose an alternative)

draft-dekater-scion-controlplane.md
```diff
@@ -1139,6 +1139,12 @@ Depending on the selection criteria, it may be necessary to keep more candidate
 
 - The *propagation interval* SHOULD be at least "5" (seconds) for intra-ISD beaconing and at least "60" (seconds) for core beaconing.
 
+Note that under certain conditions, to ensure quick connectivity establishment, an AS MAY attempt to forward a PCB more frequently ("fast recovery"):
+
+- if no beacon was successfully sent over the interface in the current propagation interval (e.g. because the corresponding RPC failed)
+
+- or if the AS receives a previously unknown path segment
+
```
nicorusti, line 1147: @jiceatscion @matzf @oncilla do these conditions sound reasonable? How would the AS check that it receives a previously unknown segment, shall we be more precise about this? 

ON a similar note, @tzaeschke 's  comments were: 
>  Is this subject to the "Best 5 PCB" rule? -> If a new path is immediately forwarded, does it count towards th "best 5"? If not, then we are effectively forwarding >5 paths, correct? If yes, then the first 5 paths are always the best until they expire and can be replaced with other paths?

> What means "unknown path"? Does it refer to all links in the segment or just the remote AS? If it is all links in the segment, then there may be many new paths coming in all the time that need to be forwarded immediately, or not?

oncilla: >  How would the AS check that it receives a previously unknown segment, shall we be more precise about this?

I don't think this is accurate. We do not do fast recovery if we receive a previously unknown path segment in all cases.

I think what it really boils down to is the following:

If we have not successfully distributed at least one beacon within the last beaconing interval, be it due to failing RPCs, or because no beacon was available to propagate, we enter the fast recovery mode.

> Is this subject to the "Best 5 PCB" rule? -> If a new path is immediately forwarded, does it count towards th "best 5"? If not, then we are effectively forwarding >5 paths, correct? If yes, then the first 5 paths are always the best until they expire and can be replaced with other paths?

The questions are mute, because we do not immediately beacon if a new path segment is discovered.

draft-dekater-scion-controlplane.md
```diff
@@ -1301,7 +1307,7 @@ A PCB originated by a given control service is validated by all the control serv
...
+This bias comes in addition to a structural delay: PCBs are propagated at a configurable interval (typically, around one minute). As a result of this and the way they are iteratively constructed, PCBs with N hops may be validated at worst case up to N intervals (so maximally N minutes) after origination. This creates a constraint on the expiration of hops. Hops of the minimal expiration time (337.5 seconds - see [](#hopfield)) would render useless any PCB describing a path longer than 5 hops. For this reason, it is unadvisable to create hops with a short expiration time, that should be around 6 hours.
```
jiceatscion, line 1310: "be validated at worst case up to N intervals (so maximally N minutes)". This is triply redundant:

"up to" means "<=" means "maximally" means "in the worst case". I give you that the "maximally" phrase is about duration and not intervals, but still, that was more than clear enough as it was.

draft-dekater-scion-controlplane.md
jiceatscion: Ok. so if we want to be both honest and minimalist, we should just give permission to the current implementation do do what it does, without mandating it (since I don't think we really know how good or bad it is). how about:

"
Note that to ensure quick connectivity establishment, an AS MAY attempt to forward a PCB more frequently ("fast recovery") if no attempt is know to have succeeded. Current practice is to increase the frequency of attempts if and only if no beacon has been successfully propagated during the last beaconing interval.
"
?

draft-dekater-scion-controlplane.md
```diff
@@ -1338,9 +1344,9 @@ If the same AS has 1000 child links, the propagation of the beacons will require
...
+On a cold start of the network, path segments to each AS are discovered at worst after a number of propagation steps proportional to the longest path. As mentioned, the longest path is typically not long. With a 5 second propagation period and a generous longest path of length 10, all path segments are discovered after 25 seconds on average.
 
-When a new parent-child link is added to the network, the parent AS will propagate the available PCBs in the next propagation event. If the AS on the child side of the new link is a leaf AS, path discovery is thus complete after one single propagation interval. Otherwise, child ASes at distance D below the new link, learn of the new link after D further propagation intervals.
+When a new parent-child link is added to the network, the parent AS will propagate the available PCBs at latest in the next propagation event. If the AS on the child side of the new link is a leaf AS, path discovery is thus complete after one single propagation interval. Otherwise, child ASes at distance D below the new link, might learn of the new link at worst after D further propagation intervals, if on path ASes do not use "fast recovery".
```
jiceatscion, line 1349: "...propagate the available PCBs at latest in..."

"...propagate the available PCBs at the latest in..."

nicorusti: changed

draft-dekater-scion-controlplane.md
nicorusti: ok

draft-dekater-scion-controlplane.md
nicorusti: How about this mix of your two wordings? 
```suggestion
Note that to ensure quick connectivity establishment, an AS MAY attempt to forward a PCB more frequently ("fast recovery"). Current practice is to increase the frequency of attempts if no PCB propagation is know to have succeeded within the last propagation interval:

- because the corresponding RPC failed

- or because no beacon was available to propagate
```

draft-dekater-scion-controlplane.md
```diff
@@ -1356,9 +1361,9 @@ All of these are manageable on a present day small server or desktop machine.
 For much larger, more highly connected ASes, the path-discovery tasks of the control service can be distributed over many instances in order to increase the PCB throughput.
 
 
-On a cold start of the network, full connectivity is obtained after a number of propagation steps corresponding to the diameter of the network. Assuming a network diameter of 6, this corresponds to roughly 3 minutes on average.
+On a cold start of the network, full connectivity is obtained at worst after a number of propagation steps corresponding to the diameter of the network. Assuming a network diameter of 6, this corresponds to roughly 3 minutes on average.
```
matzf, line 1364: Remove this "at worst"; can't be fewer steps (even if they'd be triggered quicker)

draft-dekater-scion-controlplane.md
```diff
@@ -1338,9 +1343,9 @@ If the same AS has 1000 child links, the propagation of the beacons will require
...
+On a cold start of the network, path segments to each AS are discovered at worst after a number of propagation steps proportional to the longest path. As mentioned, the longest path is typically not long. With a 5 second propagation period and a generous longest path of length 10, all path segments are discovered after 25 seconds on average.
 
-When a new parent-child link is added to the network, the parent AS will propagate the available PCBs in the next propagation event. If the AS on the child side of the new link is a leaf AS, path discovery is thus complete after one single propagation interval. Otherwise, child ASes at distance D below the new link, learn of the new link after D further propagation intervals.
+When a new parent-child link is added to the network, the parent AS will propagate the available PCBs at the latest in the next propagation event. If the AS on the child side of the new link is a leaf AS, path discovery is thus complete after one single propagation interval. Otherwise, child ASes at distance D below the new link, might learn of the new link at worst after D further propagation intervals, if on path ASes do not use "fast recovery".
```
matzf, line 1348:
```suggestion
When a new parent-child link is added to the network, the parent AS will propagate the available PCBs in the next propagation event. If the AS on the child side of the new link is a leaf AS, path discovery is thus complete after at most one propagation interval. Otherwise, child ASes at distance D below the new link, learn of the new link after at worst D further propagation intervals.
```

"fast recovery" does not accelerate finding _additional_ paths. It could only help if the AS and its entire subtree was previously disconnected. That seems an unnecessary weird edge case to explain, I'd just leave it out (and it remains correct).

draft-dekater-scion-controlplane.md
```diff
@@ -1316,10 +1321,10 @@ The path discovery mechanism balances the number of discovered paths and the tim
...
 
-Interesting metrics for scalability and speed of path discovery are the time until all discoverable path segments have been discovered after a "cold boot", and the time until new link is usable.
-Generally, the time until a specific PCB is built depends on its length and the propagation interval.
-At each AS, the PCB will be processed and propagated at the subsequent propagation event. As propagation events are not synchronized between different ASes, a PCB arrives at a random point in time during the interval and is buffered before potentially being propagated.
-With a propagation interval T at each AS, the mean time until the PCB is propagated in one AS therefore is T / 2 and the mean total time for the propagation steps of a PCB of length L is L * T / 2 (with a variance of L * T^2 / 12).
+Interesting metrics for scalability and speed of path discovery are the time until all discoverable path segments have been discovered after a "cold start", and the time until new link is usable.
+Generally, the time until a specific PCB is built depends on its length, the propagation interval, wether on-path ASes use "fast recovery".
```
matzf, line 1325: wether -> whether

The fast recovery mechanism only significantly affects the path discovery time for (the first) down segments.
In core beaconing, it's mostly irrelevant for establishing connectivity, as it is disabled as soon as one PCB was propagated over an interface. Because origination/propagation starts from everywhere, this does not create any meaningful connectivity.

If you want to mention it at all, I would describe it specifically for the down beaconing case, at the end of the paragraph starting with "On a cold start of the network ...". Something like:
"When all ASes start propagation interval just after they've received the first PCBs from any of their providers (see 'fast recovery'), the construction of a first path to connect each AS to the ISD core is accelerated.".

draft-dekater-scion-controlplane.md
nicorusti: right, I'll do

draft-dekater-scion-controlplane.md
nicorusti: Thanks, I added your wording there.

nicorusti:
Disucssed with JC, looks good enough to merge.

PR #53, opened 2024-07-18
wormhole attack: mention proof of transit as detection mechansim

### LINKED ISSUE #52: Wormhole attack: mention proof of transit as a defense

nicorusti:
Resolves #52


draft-dekater-scion-controlplane.md
```diff
@@ -1665,7 +1665,7 @@ A malicious AS M1 can send a PCB not only to their downstream neighbor ASes, but
...
 
-To defend against this attack, methods to detect the wormhole attack are needed.  Per link or path latency measurements can help reveal the wormhole and render the fake peering link suspicious or unattractive. Without specific detection mechanisms these so-called wormhole attacks are unavoidable in routing.
+To defend against this attack, methods to detect the wormhole attack are needed. Per link or path latency measurements can help reveal the wormhole and render the fake peering link suspicious or unattractive. Proof of transit mechanisms can also reveal such attack. Without specific detection mechanisms these so-called wormhole attacks are unavoidable in routing.
```
jiceatscion, line 1668: I am really no security expert, so my reading is worth the same as some random non-expert in your audience. From that naive perspective, I fail to grasp how recipients or senders requiring a proof of transit would help detect the wormhole. After all, the two attacker ASes are colluding to rewrite the packets. It seems to me they ought to have everything they need to write the fake proof of transit as well; the peering link could be real. In fact, there is no need to rewrite the packets; they could be tunneled into honest scion traffic taking the long route between the two attackers, making the peering link basically real. But that's probably just me not understanding how the proof is constructed.

I am also not quite sure that the scenario qualifies as a wormhole attack. The usual scenario is that traffic is actually taken on shorter route without permission. So, the attacker is already on the path; the goal being to deceive the sender and receiver about their proximity (and, among other things, the perenity of the connectivity). In that case, proof of transit could defend against the attack. However, that isn't the case in your example. Your example is more of a reverse wormhole attack: an attractive wormhole is advertised but does not exist. The traffic takes a long route. So the purpose of the attacker is to insert itself on the path.

nicorusti: What I want to say is that If traffic is "swallowed" via a wormhole shortcut, then packets will skip some of the ASes on path. If there is proof of transit, then the proof for the skipped ASes would be missing, revealing the attack.
On the other hand, it is impossible to prove that packets went though other ASes.. PoT can only prove that traffic did NOT go through the intended ASes. Does it make sense?

I tried to update wording

nicorusti:
You’re right, the scenario in the draft refers to wormhole in the control plane (where PCBs are used to advertise a wormholed path), and PoT is useless there. 

I was thinking about a wormhole in the data plane. IN this case, you are also right, this is a pretty useless attack, as it requires two malicious on-path ASes. Right now, we talk about this in the DP draft `5.2.1.  Modification of the Path Header` . I’ll therefore discard this PR and open one in the CP: https://github.com/scionassociation/scion-dp_I-D/pull/34

PR #62, opened 2024-09-11
grpc: describe empty responses and mention the existence of gRPC errors

### LINKED ISSUE #61: beaconing: the grpc protobufs are described, but not the errors
Eventhough the Beacon() call does not admit any response, it can still return errors.
A grpc call signatue is (in Go syntax) "func Call(request) Response, error". So we need to describe error.
We should probably check other APIs for the same oversight.

jiceatscion:
Fixes #61


nicorusti (approved):
LGTM

PR #63, opened 2024-09-20
mtu: add clarifications.

### LINKED ISSUE #57: MTU clarifications
> Feedback Harald Alvestrand
> The description of SCION refers to using UDP/IP as the internal substrate inside an AS, and as a framing protocol on inter-AS links. This means that the Internet MTU of 1500 bytes (1200 for IPv6) will be in effect.
Path hops take 12 bytes + a couple of “segment headers” of 16 bytes - which means that a 10-hop, 3-segment path will take 168 bytes for the routing information. This is enough to reduce the MTU significantly, and the MTU will be unpredictable and maybe variable, since the path for a given packet may change dynamically.
(The packet format, like UDP’s, permits 68K-byte packets. However, fragmentation is A Thing.)

Also, I find this text is not clear: 
```
 mtu: The size of the maximum transmission unit (MTU) within the
      current AS's network.
```
Shoudln't this be the minimum MTU among alll intra-domain links in the AS (so the maximum supported MTU for intra-domain forwarding)? Maybe we can find a better wording

Related issues: 
* DP draft: https://github.com/scionassociation/scion-dp_I-D/issues/39
* Implementations: https://github.com/scionproto/scion/issues/2608

jiceatscion:
Fixes #57

### DIFF
--- a/draft-dekater-scion-controlplane.md
+++ b/draft-dekater-scion-controlplane.md
@@ -884,7 +884,7 @@ The following code block defines the signed body of one AS entry in Protobuf mes
 - `next_isd_as`: The ISD-AS number of the downstream AS to which the PCB SHOULD be forwarded.
 - `hop_entry`: The hop entry (`HopEntry`) with the information required to forward this PCB through the current AS to the next AS. This information is used in the data plane. For a specification of the hop entry, see [](#hopentry).
 - `peer_entries`: The list of optional peer entries (`PeerEntry`). For a specification of one peer entry, see [](#peerentry).
-- `mtu`: The size of the maximum transmission unit (MTU) within the current AS's network.
+- `mtu`: The maximum transmission unit (MTU) that is supported by all hosts within the AS being described. This is set by the control service adding the entry to the beacon. How the control service obtains this information is implementation dependent. Current practice is do make it a configuration item.
 - `extensions`: List of (signed) extensions (optional). PCB extensions defined here are part of the signed AS entry. This field SHOULD therefore only contain extensions that include important metadata for which cryptographic protection is required. For more information on PCB extensions, see [](#pcb-ext).
 
 
@@ -943,8 +943,9 @@ The following code block defines the hop entry component `HopEntry` in Protobuf 
 ~~~~
 
 - `hop_field`: Contains the authenticated information about the ingress and egress interfaces in the direction of beaconing. The data plane needs this information to forward packets through the current AS. For further specifications, see [](#hopfield).
-- `ingress_mtu`: Specifies the maximum transmission unit (MTU) of the ingress interface of the current AS.
+- `ingress_mtu`: Specifies the maximum transmission unit (MTU) of the ingress interface (in beaconing direction) of the hop being described. Packets traversing the associated inter-AS link in either direction cannot be larger than this size. As a result, no path segment constructed from the containing beacon can have an MTU greater than this. How the control service obtains the MTU of an inter-AS link is implementation dependent. It may be discovered or configured. Current practice to make it a configuration item.
 
+In this description, MTU and packet size are to be understood in the same sense as in {{RFC1122}}. That is, exclusive of any layer 2 framing or packet encapsulation (for links using an underlay network).
 
 #### Hop Field {#hopfield}
 
@@ -1014,8 +1015,11 @@ The following code block defines the peer entry component `PeerEntry` in Protobu
 
 - `peer_isd_as`: The ISD-AS number of the peer AS. This number is used to match peering segments during path construction.
 - `peer_interface`: The 16-bit interface identifier of the peering link on the peer AS side. This identifier is used to match peering segments during path construction.
-- `peer_mtu`: Specifies the maximum transmission unit MTU on the peering link.
+- `peer_mtu`: Specifies the maximum transmission unit (MTU) of the peering link being described. Packets traversing this link in either direction cannot exceed this size. As a result, no path segment constructed from the containing beacon can have an MTU greater than this. How the control service obtains the MTU of an inter-AS link is implementation dependent. It may be discovered or configured. Current practice to make it a configuration item.
 - `hop_field`: Contains the authenticated information about the ingress and egress interfaces in the current AS (coming from the peering link, in the direction of beaconing - see also {{figure-6}}). The data plane needs this information to forward packets through the current AS. For further specifications, see [](#hopfield).
+
+In this description, MTU and packet size are to be understood in the same sense as in {{RFC1122}}. That is, exclusive of any layer 2 framing or packet encapsulation (for links using an underlay network).
+
 
 ~~~~
    +-----------+
@@ -1447,6 +1451,18 @@ The control service of a non-core AS has to register the newly created down-segm
 - `map<int32, Segments> segments`: Represents a separate list of segments for each path segment type. The key is the integer representation of the corresponding `SegmentType`.
 - `SegmentRegistrationResponse`: an empty message returned as an acknowledgement upon success.
 
+## Path MTU {#path-mtu}
+
+Paths represent a sequence of ASes and inter-AS links; each with possibly different MTUs. As a result, a path has an effective MTU which is the least of the MTUs of each link and ASes it traverses. The relevant MTUs for calculating a path's effective MTU are available from the segments used in its construction:
+
+* The MTU of the first and last ASes (represented by the mtu field of the corresponding [AS Entries](#ase-sign))
+* The MTU of each inter-AS link or peering link (indicated by the ingress_mtu field of each [](#hopentry) or the peer_mtu field of each [](#peerentry) used)
+* The MTU of any intra-AS network traversed if the ingress and egress interfaces of a hop exist on two different border routers
+
+Regarding the later point: A SCION host using a path segment has no means of knowing whether a given hop implies traversing the corresponding AS' internal network or not, nor whether a possibly larger MTU applies. In addition, a SCION control plane implementation is NOT REQUIRED to account for this in the hop's ingress_mtu. As a result, a scion host building a path must assume that all hops are additionally constrained by the internal network MTU of each AS traversed.
+
+This could be optimized if the control plane implementations made the hop's ingress_mtu field no greater than the MTU of the path between the two border routers involved in that hop. However existing implementations have not historically done this.
+
 # Path Lookup {#lookup}
 
 The *path lookup* is a fundamental building block of SCION's path management, as it enables endpoints to obtain path segments found during path exploration and registered during path registration. This allows the endpoints to construct end-to-end paths from the set of possible path segments returned by the path lookup process. The lookup of paths still happens in the control plane, whereas the construction of the actual end-to-end paths happens in the data plane.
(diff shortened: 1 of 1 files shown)


draft-dekater-scion-controlplane.md
```diff
@@ -884,7 +884,7 @@ The following code block defines the signed body of one AS entry in Protobuf mes
 - `next_isd_as`: The ISD-AS number of the downstream AS to which the PCB SHOULD be forwarded.
 - `hop_entry`: The hop entry (`HopEntry`) with the information required to forward this PCB through the current AS to the next AS. This information is used in the data plane. For a specification of the hop entry, see [](#hopentry).
 - `peer_entries`: The list of optional peer entries (`PeerEntry`). For a specification of one peer entry, see [](#peerentry).
-- `mtu`: The size of the maximum transmission unit (MTU) within the current AS's network.
+- `mtu`: The maximum transmission unit (MTU) that is supported by all hosts within the AS being described. This is set by the control service adding the entry to the beacon. How the control service obtains this information is implementation dependent. Current practice is do make it a configuration item.
```
nicorusti, line 887:
```suggestion
- `mtu`: The maximum transmission unit (MTU) that is supported by all intra-domain links within the current AS. This value is set by the control service when adding the AS entry to the beacon. How the control service obtains this information is implementation dependent. Current practice is to make it a configuration item.
```
I think it is better to refer to the intra-domain underlay rather than to hosts

draft-dekater-scion-controlplane.md
```diff
@@ -943,8 +943,9 @@ The following code block defines the hop entry component `HopEntry` in Protobuf
 ~~~~
 
 - `hop_field`: Contains the authenticated information about the ingress and egress interfaces in the direction of beaconing. The data plane needs this information to forward packets through the current AS. For further specifications, see [](#hopfield).
-- `ingress_mtu`: Specifies the maximum transmission unit (MTU) of the ingress interface of the current AS.
+- `ingress_mtu`: Specifies the maximum transmission unit (MTU) of the ingress interface (in beaconing direction) of the hop being described. Packets traversing the associated inter-AS link in either direction cannot be larger than this size. As a result, no path segment constructed from the containing beacon can have an MTU greater than this. How the control service obtains the MTU of an inter-AS link is implementation dependent. It may be discovered or configured. Current practice to make it a configuration item.
```
nicorusti, line 946:
```suggestion
- `ingress_mtu`: Specifies the maximum transmission unit (MTU) of the ingress interface (in beaconing direction) of the hop being described. As a result, path segments constructed from the containing beacon MUST NOT have an MTU greater than this. How the control service obtains the MTU of an inter-AS link is implementation dependent. It may be discovered or configured. Current practice to make it a configuration item.
```

draft-dekater-scion-controlplane.md
```diff
@@ -943,8 +943,9 @@ The following code block defines the hop entry component `HopEntry` in Protobuf
...
 
+In this description, MTU and packet size are to be understood in the same sense as in {{RFC1122}}. That is, exclusive of any layer 2 framing or packet encapsulation (for links using an underlay network).
```
nicorusti, line 948: Are you referring to a specific section of RFC1122? I skimmed through it and I did not find where you are referring to immediately. In the terminology section, it only says: 
```
         MTU
              The maximum transmission unit, i.e., the size of the
              largest packet that can be transmitted.
```

draft-dekater-scion-controlplane.md
```diff
@@ -1014,9 +1015,12 @@ The following code block defines the peer entry component `PeerEntry` in Protobu
 
 - `peer_isd_as`: The ISD-AS number of the peer AS. This number is used to match peering segments during path construction.
 - `peer_interface`: The 16-bit interface identifier of the peering link on the peer AS side. This identifier is used to match peering segments during path construction.
-- `peer_mtu`: Specifies the maximum transmission unit MTU on the peering link.
+- `peer_mtu`: Specifies the maximum transmission unit (MTU) of the peering link being described. Packets traversing this link in either direction cannot exceed this size. As a result, no path segment constructed from the containing beacon can have an MTU greater than this. How the control service obtains the MTU of an inter-AS link is implementation dependent. It may be discovered or configured. Current practice to make it a configuration item.
```
nicorusti, line 1018:
```suggestion
- `peer_mtu`: Specifies the maximum transmission unit (MTU) of the peering link being described. As a result, path segments constructed from the containing beacon MUST NOT have an MTU greater than this. How the control service obtains the MTU of an inter-AS link is implementation dependent. It may be discovered or configured. Current practice to make it a configuration item.
```

draft-dekater-scion-controlplane.md
```diff
@@ -1447,6 +1451,18 @@ The control service of a non-core AS has to register the newly created down-segm
...
+
+* The MTU of the first and last ASes (represented by the mtu field of the corresponding [AS Entries](#ase-sign))
+* The MTU of each inter-AS link or peering link (indicated by the ingress_mtu field of each [](#hopentry) or the peer_mtu field of each [](#peerentry) used)
+* The MTU of any intra-AS network traversed if the ingress and egress interfaces of a hop exist on two different border routers
+
+Regarding the later point: A SCION host using a path segment has no means of knowing whether a given hop implies traversing the corresponding AS' internal network or not, nor whether a possibly larger MTU applies. In addition, a SCION control plane implementation is NOT REQUIRED to account for this in the hop's ingress_mtu. As a result, a scion host building a path must assume that all hops are additionally constrained by the internal network MTU of each AS traversed.
+
+This could be optimized if the control plane implementations made the hop's ingress_mtu field no greater than the MTU of the path between the two border routers involved in that hop. However existing implementations have not historically done this.
```
nicorusti, line 1465:
```suggestion
This may be optimized if the control plane implementations made the hop's `ingress_mtu` field no greater than the MTU of the path between the two border routers involved in that hop. However existing implementations have not historically done this in favor of the more conservative approach to use the maximum AS MTU specified in [](#ase-sign).
```

draft-dekater-scion-controlplane.md
```diff
@@ -1082,6 +1086,7 @@ For the purpose of constructing and propagating path segments, an AS control ser
 - Neighbor ISD-AS number
 - Neighbor interface underlay address
 
+In addition, the maximum MTU supported by all intra-AS links MAY be configured.
```
nicorusti, line 1089: JFYI I've added this

nicorusti (commented):
Thanks a lot, looks good. I proposed rewording of some sections and added a note in the config section

draft-dekater-scion-controlplane.md
jiceatscion: I think neither my nor your formulation are accurate. A path segment does not have an end-to-end MTU. Only a path does. In addition, we cannot say "MUST" or "MUST NOT" as no-one is setting any end-2-end MTU for a path at the protocol level. It is ony a property that results from the various MTU fields. How about:

"
- `ingress_mtu`: Specifies the maximum transmission unit (MTU) of the ingress interface (in beaconing direction) of the hop being described. The MTU of path segments constructed from the containing beacon is necessarily less than or equal to this value. How the control service obtains the MTU of an inter-AS link is implementation dependent. It may be discovered or configured. Current practice to make it a configuration item.
"

draft-dekater-scion-controlplane.md
jiceatscion: A precise reference would be complicated. I was shocked to discover how poorly the term is defined. This is the closest I found to a definition and it is spread between the paragraph you quote and the text after it:

> The terms frame, packet, datagram, message, and segment are
>          illustrated by the following schematic diagrams:
> 
>          A. Transmission on connected network:
>            _______________________________________________
>           | LL hdr | IP hdr |         (data)              |
>           |________|________|_____________________________|
> 
>            <---------- Frame ----------------------------->
>                     <----------Packet -------------------->

So, the MTU is defined first as the property of a packet, and then a packet is defined as the portion of the Frame that follows the LL header.

draft-dekater-scion-controlplane.md
jiceatscion: Same response as before and same suggestion for a third brew of the text.

draft-dekater-scion-controlplane.md
jiceatscion: No, that isn't quite true. The CS implementations do not conservatively use the AS's MTU as a limit on the ingress MTU. They let the endpoints do that. As a result, the endpoints *have* to do it. As a result, there's no point for the CS to offer the optimal MTU for a given hop.

The truth is that we'd need a migration strategy:
* Phase 1: reduce the ingress MTU to that of the specific HOP (which has no effect on consumers of the HOP).
* Phase 2: wait for all CS'es in the world to run running that modified code.
* Phase 3: update the daemon and other segment consuming code to rely on the ingress MTU only and consider the AS's MTU of only the last HOP of the path.

How the heck do I summarize that in a normative doc? Yet, if we don't say it, someone will object and say the protocol is wasting a performance opportunity.

draft-dekater-scion-controlplane.md
jiceatscion: ack.

draft-dekater-scion-controlplane.md
nicorusti: True, I added your formulation.

draft-dekater-scion-controlplane.md
nicorusti: I updated my suggestion above. As this is a peering link, I think we cannot use the same wording as for regular links (we cannot refer to the MTU in a path segment, as the peering link intrinsically merges two distinct segments). Let me know wat you think about this wording: 
> The MTU of paths via such link is necessarily less than or equal to this value.

draft-dekater-scion-controlplane.md
jiceatscion: And I changed "MTU of path segments" to "MTU of paths", since we agreed that the MTU of a segment is undefined.

draft-dekater-scion-controlplane.md
jiceatscion: Yes, that's what I would have done. I did apply the same change to the ingress MTU for a similar reason: there is no segment MTU; it all depends on which portion you use.

draft-dekater-scion-controlplane.md
nicorusti: Thank you for the clarification, I see your point. Let me propose another wording (I have to open a new thread).

draft-dekater-scion-controlplane.md
nicorusti: Ok, then let's perhaps leave the reference as it is, I think the text is clear enough,

draft-dekater-scion-controlplane.md
```diff
@@ -1447,6 +1452,20 @@ The control service of a non-core AS has to register the newly created down-segm
...
+* The MTU of each inter-AS link or peering link (indicated by the ingress_mtu field of each [](#hopentry) or the peer_mtu field of each [](#peerentry) used)
+* The MTU of any intra-AS network traversed if the ingress and egress interfaces of a hop exist on two different border routers
+
+It is then made available to endpoints during the path lookup process.
+
+Regarding the later point: A SCION endpoint using a path segment has no means of knowing whether a given hop implies traversing the corresponding AS' internal network or not, nor whether a possibly larger MTU applies. In addition, SCION control plane implementations are NOT REQUIRED to account for this in the hop's `ingress_mtu` field. As a result, a SCION endpoint looking up a path MUST assume that all hops are additionally constrained by the internal network MTU of each AS traversed.
+
+This could be optimized if the control plane implementations made the hop's ingress_mtu field no greater than the MTU of the path between the two border routers involved in that hop. However existing implementations have not historically done this.
```
nicorusti, line 1467:
```suggestion
* The MTU of each intra-AS network traversed (represented by the MTU field of the corresponding [AS Entries](#ase-sign))
* The MTU of each inter-AS link or peering link (indicated by the ingress_mtu field of each [](#hopentry) or the peer_mtu field of each [](#peerentry) used)


Such information is then made available to endpoints during the path lookup process (See [](#lookup)). SCION endpoints are oblivious to the topology of intermediate ASes, therefore when looking up a path they MUST assume that all hops are constrained by the intra-AS MTU of each AS traversed.
```

nicorusti (approved):

Issue #50, opened 2024-07-09
Anapaya review


nicorusti:
Following the first wave of ISE feedback, we did quite some work, especially on the control plane draft:

Please review (sorted by urgency): 
- #63 
- 2.3.3.  Effects of Clock Inaccuracy
-  2.4.  Path Discovery Time and Scalability
-  5.  Security Considerations
- Control Service gRPC API (appendix)

Other pointers:
- [Changelog](https://github.com/scionassociation/scion-cp_I-D/releases/tag/draft-dekater-scion-controlplane-04)
- [Datatracker diff v-03 to main](https://author-tools.ietf.org/api/iddiff?doc_1=draft-dekater-scion-controlplane-03&url_2=https://scionassociation.github.io/scion-cp_I-D/draft-dekater-scion-controlplane.txt) 

@shitz @oncilla @sgmonroy could one of you have a look at the overall changes and provide some feedback? If you feel, you can  directly open a PR.  We are planning to have a new submission when datatracker reopens on 21.07, before our panrg session to address #48 , and it would be great if we could also incorporate some of your feedback.

oncilla:
> The expiration of a SCION AS certificate typically ranges from	
 	   3h to 5 years.
 	   
 
Where does this number come from. Most typically it is 3days, and in some very special cases we have 5 days.
I have never seen 5 years outside of some private ISDs

oncilla:
>  Once	
 	   this number grows above the maximum recommended best PCBs set size of	
 	   50, 
 	   
Where does this number come from? 
Default is 20. AFAIK:
https://github.com/scionproto/scion/blob/aa917bb458d6675efc55b4a23fda2c694094553e/control/beacon/policy.go#L43

nicorusti:
> > The expiration of a SCION AS certificate typically ranges from
> > 3h to 5 years.
> 
> Where does this number come from. Most typically it is 3days, and in some very special cases we have 5 days. I have never seen 5 years outside of some private ISDs

This comes from https://github.com/scionassociation/scion-cp_I-D/commit/297e8d2c828228e22542cd27c6b708f29586a440  @jiceatscion do you have more background?

@oncilla you're right, in the PKI draft we write `The RECOMMENDED maximum validity period of a CP AS certificate is: 3 days.` I don't think, however, we set an upper bound to the validity. 

I see two action items: 

- [ ] Correct CP draft to say "3h to 5 _days_".
[...]

PR #69, opened 2024-10-08
cp_draft: fix recommended/typical values for best PCB set size and AS cert TTL

jiceatscion:
Addresses #50


nicorusti:
Can you add one line to the change log?

jiceatscion:
If you update the branch from main, there'll be a change log to update.

PR #80, opened 2024-12-20
Address Til's comments.

### LINKED ISSUE #46: Clarifications in introduction and beaconing (Til's feedback)
Copied over from feeedbac received by @tzaeschke  in https://github.com/scionassociation/scion-cp_I-D/pull/42#issuecomment-2211084556

# Introduction
## Avoiding Circular Dependencies and Partitioning
- Does this section (title) make sense as it is? The two topics (circular dependencies and partitioning) appear unrelated. Partitioning is discussed in a separate subsection
- Also, this section claims to contain a list that explains how circular dependencies are avoided.
  However, I am not sure how anything in the list explains anything about circular dependencies?

### Partition and Healing
- `ASes could always switch to otherwise unused links`. 

  - What are `unused` links? `Unused` indicates that they are not in the "Best PCB" set. 
  If that is the case, then we cannot simply "switch" to use them, we first need to have them discovered by beacons. This takes time, see `propagation interval` (unless PCB are forwarded immediately) and beacon `origination interval`.
  - Also: Does "Healing" include "adding new links"? In this case we also need to wait for the `propagation interval` and the `origination interval`.
  As I understand, with propagation interval being set to e.g. 10Minutes, this adds up to 5hops*1/2*10min= 25minutes for 5 hops.

# Path Exploration or Beaconing {#beaconing}
## Introduction and Overview

- `Intra-ISD beaconing`: Isn't this incomplete? How are DOWN segments created? I think leaf ASes need to propagate paths back to the COREs? -> Link to #intra-reg ?
- `Inter-ISD beaconing`: Similarily, I think there is a step where CORE beacons share their path DB with other CORE ASes in the same ISD?

### Extending a PCB
- `selects the best combinations`: Maybe link to a section that explains how this works? And what "best" means?

### PCB Validity
- `For the purpose of validation, a timestamp is considered "future" if it is later than the current time at the point of validation plus the minimum expiration time of a hop field (337.5 seconds, see ).`
Maybe add an explanation why we add the minimum expiration time here?
Shouldn't "future" simply be timestamp + some_delta_to_account_for_server_time_inaccuracies, where the delta is maybe a few seconds rather than 5.5 minutes?

#### Storing and Selecting Candidate PCBs

- `temporary storage`: Maybe clarify what "temporary" means. 
  - How long are PCBs stored? 
  - Under what circumstances are they removed, e.g. when they expire? 
  - Or possibly replaced with a new version? What if the new version has an earlier expiration date?

- `At each propagation event, each AS selects a set of the best PCBs from the candidates in the beacon store`

- `The best PCBs set size SHOULD be at most "50" (PCBs) for intra-ISD beaconing and at most "5" (PCBs) for core beaconing.`
  - I found this a bit confusing: the "50" appears to be the total number of beacons forwarded for non-core ASes, whereas the "5" refers to the number of 
    PCBs _per_ remote CORE AS. Maybe clarify this?

- `Note that during bootstrapping and if the AS obtains a PCB containing a previously unknown path, the AS SHOULD forward the PCB immediately, [...].`
  - Is this true? It appears to conflict with many other parts of the document that talk about propagation intervals in the contect of cold-start.
  - Is there a difference between "bootstrapping" (used here) and "cold-start" (used in other places)? Maybe stick to one term or explicitly declare equality?
  - Is this subject to the "Best 5 PCB" rule? -> If a new path is immediately forwarded, does it count towards th "best 5"? If not, then we are effectively
    forwarding >5 paths, correct? If yes, then the first 5 paths are always the best until they expire and can be replaced with other paths? 
  - What means "unknown path"? Does it refer to all links in the segment or just the remote AS? If it is all links in the segment, then there may be many new paths
    coming in all the time that need to be forwarded immediately, or not?

jiceatscion:
Fixes #46
Draft [diff with main](https://author-tools.ietf.org/api/iddiff?url_1=https://scionassociation.github.io/scion-cp_I-D/draft-dekater-scion-controlplane.txt&url_2=https://scionassociation.github.io/scion-cp_I-D/til_feedback/draft-dekater-scion-controlplane.txt)


nicorusti:
Thanks a lot, I'll review after holidays!

draft-dekater-scion-controlplane.md
```diff
@@ -425,11 +427,12 @@ In SCION, the *Control Service* of each AS is responsible for the beaconing proc
...
-
-- *Intra-ISD beaconing* is the process whereby the Control Service of a core AS initiates PCBs and sends them at regular intervals to non-core child ASes (typically customer ASes) within the same ISD. The Control Service of a non-core child AS receives these PCBs and forwards them to its child ASes, and so on until the PCB reaches an AS without any customer (leaf AS). As a result, all ASes within an ISD receive path segments that enable them to reach the core ASes within the same ISD.
+- *Intra-ISD beaconing* creates path segments from core ASes to non-core ASes. For this, the Control Services of core ASes create PCBs and sends them to the non-core child ASes (typically customer ASes) at regular intervals. The Control Service of a non-core child AS receives these PCBs and forwards them to its child ASes, and so on until the PCB reaches an AS without any customer (leaf AS). As a result, all ASes within an ISD receive path segments to reach the core ASes of their ISD  and register reciprocal segments with the Control Service of the associated core ASes.
```
jiceatscion, line 430: without any customer -> without any children

draft-dekater-scion-controlplane.md
```diff
@@ -386,18 +386,22 @@ SCION endpoints use wildcard AS `0:0:0` to designate any core AS, e.g. to place
...
+Besides inter-dependencies, another threat to the Internet is network partitionig which occurs when one network is split into two because of a link failure. Partitioning of the global SCION inter-domain network is much less likely to happen as during normal operation the full network fabric is available, offering multiple paths between all ASes. Even during failures there is no special failure mode required as SCION-enabled ASes can always switch to already known paths that use other links.
 
 Recovering from a partitioned network is also seamless as only coarse time synchronization between the partitions is required to resume normal operation and move forward with updates of the cryptographic material. [](#clock-inaccuracy) further describes the impact of time synchronization.
```
nicorusti, line 406: I think referencing this section would address Til's comment: 
> Also: Does "Healing" include "adding new links"? In this case we also need to wait for the propagation interval and the origination interval.
As I understand, with propagation interval being set to e.g. 10Minutes, this adds up to 5hops1/210min= 25minutes for 5 hops.

```suggestion
Recovering from a partitioned network is also seamless as only coarse time synchronization between the partitions is required to resume normal operation and move forward with updates of the cryptographic material. [](#clock-inaccuracy) further describes the impact of time synchronization and [](#clock-inaccuracy) the impact of path discovery time.
```

Also, we don't set an upper bound to propagation interval. Shall we?

draft-dekater-scion-controlplane.md
```diff
@@ -424,11 +428,12 @@ In SCION, the *Control Service* of each AS is responsible for the beaconing proc
...
+- *Intra-ISD beaconing* creates path segments from core ASes to non-core ASes. For this, the Control Services of core ASes create PCBs and sends them to the non-core child ASes (typically customer ASes) at regular intervals. The Control Service of a non-core child AS receives these PCBs and forwards them to its child ASes, and so on until the PCB reaches an AS without any children (leaf AS). As a result, all ASes within an ISD receive path segments to reach the core ASes of their ISD  and register reciprocal segments with the Control Service of the associated core ASes.
 
 On its way, a PCB accumulates cryptographically protected path and forwarding information per traversed AS. At every AS, metadata as well as information about the AS's ingress and egress interfaces is added to the PCB. The full PCB message format is described in [](#pcbs).
 
+The process of using PCBs to construct segments is described in details in [](#path-segment-reg)
```
nicorusti, line 435:
```suggestion
On its way, a PCB accumulates cryptographically protected path and forwarding information per traversed AS. At every AS, metadata as well as information about the AS's ingress and egress interfaces is added to the PCB. The full PCB message format is described in [](#pcbs). PCBs are used to construct path segments. ASes register them to make them available to other ASes, as described in [](#path-segment-reg).
```

draft-dekater-scion-controlplane.md
```diff
@@ -1090,7 +1095,7 @@ To be valid (that is, usable to construct a valid path), a PCB MUST:
 * Have a timestamp ([](#seginfo)) that is not in the future.
 * Contain only unexpired hops ([](#hopfield)).
 
-For the purpose of validation, a timestamp is considered "future" if it is later than the current time at the point of validation plus the minimum expiration time of a Hop Field (337.5 seconds, see [](#hopfield)).
+For the purpose of validation, a timestamp is considered "future" if it is later than the current time at the point of validation plus an allowance for differences between local and origination clock. As an allowance, it is suggested to use the granularity of the hopfield expiration time (that is 337.5 seconds, see [](#hopfield)).
```
nicorusti, line 1098:
```suggestion
For the purpose of validation, a timestamp is considered "future" if it is later than the current time at the point of validation plus an allowance for differences between the validator's and originator's clock. As an allowance, it is recommended to use the granularity of the hopfield expiration time (that is 337.5 seconds, see [](#hopfield)).
```

nicorusti (changes requested):
@jiceatscion I left a few comments, can you have a look? W can then discuss

@tzaeschke , feel free to also have a look , we are trying to address your remaining points from #46

draft-dekater-scion-controlplane.md
```diff
@@ -386,18 +386,22 @@ SCION endpoints use wildcard AS `0:0:0` to designate any core AS, e.g. to place
...
 - Path reversal: In SCION, every path is reversible. That is, the receiver of a packet can reverse the path in the packet header in order to produce a reply packet without having to perform a path lookup. Such a packet follows the original packet's path backwards.
-- Availability of certificates: In SCION, every entity is required to be in possession of all cryptographic material (including the ISD's TRC and certificates) that is needed to verify any message it sends. This (together with the path reversal) means that the receiver of a message can always obtain all this material by contacting the sender.<br>
+- Availability of certificates: In SCION, every entity is required to be in possession of all cryptographic material (including the ISD's TRC and certificates) that is needed to verify any message it sends. This (together with the path reversal) means that the receiver of a message can always obtain all this material by contacting the sender. This avoids circular dependencies between the PKI and connectivity.<br>
```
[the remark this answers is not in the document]
jiceatscion: Explaining bootstraping flows is always hard...
The statement about knowing only immediate neighbors is absolutely true.
The path to a core AS comes from beaconing; which is enabled by having connectivity with only the immediate parent and children ASes. One-hop paths can be crafted without any prior communication. If beaconing works, then all ASes learn paths to core ASes; which they can then query to obtain segments to construct all other paths.

draft-dekater-scion-controlplane.md
jiceatscion: Did you refer to the same clock-inaccuracy sectoin twice in a row, in your suggestion?

draft-dekater-scion-controlplane.md
```diff
@@ -387,7 +387,7 @@ The text representation of SCION addresses MUST be `<ISD>-<AS>`, where `<ISD>` i
 
 For example, the text representation of AS number 65551 (0x1000f) in ISD number 4 is `4-0000:1:f`.
 
-## Avoiding Circular Dependencies and Partitioning
+## Bootstrapping ability
```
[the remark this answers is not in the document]
jiceatscion: Do you mean to replace the title "Bootstrapping" with "PKI-network circular dependency"? Why? Certificates aren't the only objects involved in the bootstrapping process. Also I suggest avoiding the phrase "circular dependencies" in section titles. It gives the impression that it's a feature. The feature is the opposite: the ability to bootstrap (or "cold-start" if that's the term we prefer).

draft-dekater-scion-controlplane.md
jiceatscion: As for propagation intervals; no I don't think so. Do I need to double check? Should we have a limit to reduce healing time? It is true that paths may take a long time to become known. That's a problem if they break often.

draft-dekater-scion-controlplane.md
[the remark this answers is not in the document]
tzaeschke: To clarify, my suggestion was to either use only one term (not both) or, otherwise explicitly state that they mean the name thing.
Just using them both without saying that they are the same is probably confusing. I think the least confusing thing would be to have only one term. This may also come up with reviewers.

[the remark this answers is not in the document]
nicorusti: Indeed, I meant that we choose one and stick to it.

In the current draft version, these are the uses: 
Bootstrap
* ## Bootstrapping ability (this chapter)
* To bootstrap the initial communication with a neighboring beacon service,
* the following bootstrapping strategies apply:
* In the changelog (irrelevant)



Cold start: 
* Interesting metrics for scalability and speed of path discovery are the time until all discoverable path segments have been discovered after a "**cold start**", and the time until new link is usable. 
* On a **cold start** of the network, path segments to each AS are discovered within a numbe
* On a **cold start** of the network, full connectivity is obtained after a number of propagation steps

Plus:
*  In the PKI draft we talk about bootstrapping of trust
* In the DP we talk about one-hop path types  "The OneHopPath path type (PathType=2) is currently used to bootstrap" 

Bootstrap wins, i will change instances of cold start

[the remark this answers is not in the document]
nicorusti: Done, resolving

draft-dekater-scion-controlplane.md
```diff
@@ -386,20 +386,24 @@ SCION endpoints use wildcard AS `0:0:0` to designate any core AS, e.g. to place
...
 
 - Neighbor-based path discovery: Path discovery in SCION is performed by the beaconing mechanism. In order to participate in this process, an AS only needs to be aware of its direct neighbors. As long as no path segments are available, communicating with the neighboring ASes is possible with the one hop path type which does not rely on any path information. SCION uses these *one hop paths* to propagate PCBs to neighboring ASes to which no forwarding path is available yet. The One Hop Path Type is described in more detail in {{I-D.dekater-scion-dataplane}}.
 - Path segment types: SCION uses different types of path segments to compose end-to-end paths. Notably, a single path segment already enables intra-ISD communication. For example, a non-core AS can reach the core of the local ISD simply by using an up segment fetched from the local path storage which is populated during the beaconing process.
```
nicorusti, line 396: @jiceatscion do you think this point is relevant to bootstrapping? I don't feel it is that much, so I think it could be removed:
```suggestion
```

draft-dekater-scion-controlplane.md
```diff
@@ -386,20 +386,24 @@ SCION endpoints use wildcard AS `0:0:0` to designate any core AS, e.g. to place
...
-- Availability of certificates: In SCION, every entity is required to be in possession of all cryptographic material (including the ISD's TRC and certificates) that is needed to verify any message it sends. This (together with the path reversal) means that the receiver of a message can always obtain all this material by contacting the sender.<br>
+- Availability of certificates: In SCION, every entity is required to be in possession of all cryptographic material (including the ISD's TRC and certificates) that is needed to verify any message it sends. This (together with the path reversal) means that the receiver of a message can always obtain all this material by contacting the sender. This avoids circular dependencies between the PKI and connectivity.<br>
+
+**Note:** For a detailed description of a TRC and more information on the availability of certificates and TRCs, see the SCION Control-Plane PKI {{I-D.dekater-scion-pki}}.
```
nicorusti, line 400: @jiceatscion how are certificates requested? Right now we reference the PKI draft, but I could not find where requesting certs is described. Could we add a better and more precise reference?

nicorusti: I guess this might be what I am referring to: https://docs.scion.org/en/latest/cryptography/interactions.html

jiceatscion: Far as I know, what you want is here: https://github.com/scionproto/scion/blob/abd39afe0a83199aca5ad59d6cc4cc383c89d4cf/proto/control_plane/v1/cppki.proto#L55
I do not know if we described the full pki rpc API anywhere. I wasn't much involved in the cppki draft.

jiceatscion: Indeed, https://docs.scion.org/en/latest/cryptography/interactions.html#messages seems to be as much doc as we have.

nicorusti: https://github.com/scionassociation/scion-cp_I-D/issues/96

nicorusti (approved):

tzaeschke (approved):

PR #84, opened 2025-01-15
This is not a standard

### LINKED ISSUE #83: Abstract - rework
Include: 
> This document contains new approaches to secure path aware networking.  However, it is not an Internet Standard, nor has it received any formal review of the IETF, nor was the work developed through the rough  consensus process.  The approaches offered in this work are offered to the community for its consideration in further evolution of the Internet.

knmeynell:
Resolves #83


nicorusti:
How do you feel about the abstract? I'm not entirely fond of the current text, perhaps it could be shortened..

nicorusti:
New abstract should mention: 
* Key function (secure path aware routing)
* Relation to other components (paths are used for forwarding by data plane)

PR #92, opened 2025-02-06
Security considerations: clarify assumptions and security properties

### LINKED ISSUE #89: Security considerations: add assumptions and security properties
- [x] Subsection 2.2.4. "Configuration” should also mention that interface identifiers must be unique within an AS
- [x] Clarify that an ISD MUST have at least one Core AS (e.g. in 2.1.  Introduction and Overview )
- [ ] Clarify if loop avoidance needs some editing (what David calls "cyclic relationships")

Add a new subsection (6.1) in security considerations: 
- [x] Assumptions (briefly summarize points in 1 below)
- [x] "Intended Security properties" (based on points in 2 below, then add a reference in the intro)

--- 
Base on feedback received from David Galan:

1. We propose that assumptions about the network and the behaviour of honest ASes are made more explicit. We would be happy to help with a specific list of assumptions:

- In our formal model, we assume that the network of physical links is such that no ISD is internally partitioned (i.e., each AS in an ISD can reach every other AS in the same ISD), --> See #46 
- every ISD has at least one core AS, and
- customer-provider relationships (and related links) form a unidirectional hierarchical topology without cyclic relationships (i.e., they form a directed acyclic graph).

An AS is called honest if its long-term keys are unknown to the attacker and it uses a unique interface identifier for each link. Let an honest path (resp., segment) be a path (resp., segment) which only traverses honest ASes.

Even though this is explicitly stated elsewhere in the draft, perhaps Subsection 2.2.4. "Configuration” should also mention that interface identifiers must be unique within an AS.

2.   Regarding correctness and security properties, we suggest inserting a new Subsection 6.1 "Correctness and Security Properties" in Section 6 "Security Considerations" containing precise descriptions in English prose that faithfully capture all the required details and avoid ambiguities.

We would be happy to contribute a list of candidate properties and their descriptions. These properties should be understood as carefully considered design goals.  Within our formal model, the properties considered so far are:
- (Connectivity) For every pair of distinct honest ASes X and Y, such that there is at least one honest path in the network of physical links from X to Y, X will eventually register enough segments to build at least one path (of any length) leading to Y.
- (Path Consistency) For every honest segment registered in any AS, its sequence of AS entries corresponds to a continuous path in the network of physical links. This property holds only if the network topology remains unchanged since the segment was first generated.
- (Loop Freedom) For every honest segment registered in any AS, its sequence of AS entries (including current and next ISD-AS, as well as interfaces) contains no duplicates.
- (Beacon Authorization) For every honest segment registered in any AS and any AS X appearing on that segment - - (except for the last one), AS X propagated a PCB corresponding to the segment prefix ending in its own entry to its successor AS on the segment.
- (Path Discoverability) For every directed honest path in the network of physical links, where all traversed ASes permit forwarding, there exists at least one protocol execution that registers a segment containing or exactly matching this path. This property holds only when the involved physical links remain stable throughout the protocol execution.

We can offer formal expressions of the properties in Linear Temporal Logic (LTL), which is part of the IEEE 1850 Standard for Property Specification Language (PSL), as well as other formal artifacts such as a labeled transition system that describes the protocol behaviors. This may be a good fit for future work (e.g. in UFMRG).

knmeynell:
Fixes #89 
Draft [diff to main](https://author-tools.ietf.org/api/iddiff?url_1=https://scionassociation.github.io/scion-cp_I-D/draft-dekater-scion-controlplane.txt&url_2=https://scionassociation.github.io/scion-cp_I-D/Security-Updates/draft-dekater-scion-controlplane.txt)

### DIFF
--- a/draft-dekater-scion-controlplane.md
+++ b/draft-dekater-scion-controlplane.md
@@ -1955,6 +1955,13 @@ This section focuses on three kinds of security risks in the control plane:
 2. When "ordinary" (non-core) adversaries try to manipulate the beaconing process (see [](#manipulate-beaconing)).
 3. Denial of Services (DoS) attacks where attackers overload different parts of the infrastructure (see [](#dos-cp)).
 
+These security considerations make assumptions about the formal model of the SCION implementation as follows:
+
+1. Every ISD has at least one core AS.
+2. The physical links within an ISD are not partitioned and each AS can reach every other AS in the same ISD.
+3. Customer-Provider relationships and their related links form a unidirectional hierarchical topology with cyclic relationships.
+
+
 ## Manipulation of the Beaconing Process by a Core Adversary {#topdown-manipulate}
 
 The first risk to the beaconing process comes from an adversary controlling one or more core ASes in an ISD. If the adversary stops all core AS(es) within an ISD from propagating PCBs, the discovery of new paths will halt. In this case, downstream ASes will notice that PCBs are no longer being propagated, but all previously discovered and still valid paths remain usable for data plane forwarding until they expire. This is an unlikely scenario, as it would require compromise of all core ASes within an ISD.
--- a/draft-dekater-scion-controlplane.md
+++ b/draft-dekater-scion-controlplane.md
@@ -206,7 +206,7 @@ Note (to be removed before publication): this document, together with the other 
 
 **Control Service**: The Control Service is the main control plane infrastructure component within a SCION AS. It is responsible for the path exploration and registration processes that take place within the Control Plane.
 
-**Core AS**: Each Isolation Domain (ISD) is administered by a set of distinguished autonomous systems (ASes) called Core ASes, which are responsible for initiating the path-discovery and -construction process (in SCION called "beaconing"). Each ISD MUST have at least one Core AS.
+**Core AS**: Each Isolation Domain (ISD) is administered by a set of distinguished autonomous systems (ASes) called core ASes, which are responsible for initiating the path-discovery and -construction process (in SCION called "beaconing"). Each ISD MUST have at least one Core AS.
 
 **Endpoint**: An endpoint is the start or the end of a SCION path, as defined in {{RFC9473}}.
 
@@ -422,7 +422,7 @@ The RPC messages are transported via the {{Connect}}'s rpc protocol; a gRPC-like
 
 In SCION, the *Control Service* of each AS is responsible for the beaconing process. The Control Service generates, receives, and propagates the *Path Segment Construction Beacons (PCBs)* on a regular basis, to iteratively construct path segments.
 
-PCBs contain topology and authentication information, and can also include additional metadata that helps with path management and selection. The beaconing process itself is divided into routing processes on two levels, where *inter-ISD* or core beaconing is based on the (selective) sending of PCBs without a defined direction, and *intra-ISD* beaconing on top-to-bottom propagation. The beaconing process is initiated by core ASes, therefore each ISD MUST have at least one core AS.
+PCBs contain topology and authentication information, and can also include additional metadata that helps with path management and selection. The beaconing process itself is divided into routing processes on two levels, where *inter-ISD* or core beaconing is based on the (selective) sending of PCBs without a defined direction, and *intra-ISD* beaconing on top-to-bottom propagation. Beaconing is initiated by core ASes, therefore each ISD MUST have at least one core AS.
 
 - *Inter-ISD or core beaconing* is the process of constructing path segments between core ASes in the same or in different ISDs. During core beaconing, the Control Service of a core AS either initiates PCBs or propagates PCBs received from neighboring core ASes to other neighboring core ASes. PCBs are periodically sent over policy compliant paths to discover multiple paths between any pair of core ASes.
 
@@ -1211,15 +1211,6 @@ Selected path segment: #------# or *------*
 ~~~~
 {: #figure-7 title="Example networks to illustrate path segment selection based on different path properties."}
 
-Some examples of path selection policies include:
-
-- BestSetSize - the recommended number of segments to propagate or register.
-- CandidateSetSize - the maximum number of segments to consider for selection.
-- MaxExpTime - the maximum value for the expiration time when extending the segment.
-- MaxHopsLength - the maximum number of hops a segment can have.
-- ASBlackList - the ASes that may not appear in a segment.
-- IsdBlackList - the ISDs that may not appear in segment.
-- AllowIsdLoop - indicates whether ISD loops should not be filtered.
 
 ### Propagation of Selected PCBs {#path-segment-prop}
 
--- a/draft-dekater-scion-controlplane.md
+++ b/draft-dekater-scion-controlplane.md
@@ -1963,18 +1963,10 @@ This section discusses three kinds of threats to the control plane:
 3. Denial of Services (DoS) attacks where attackers overload different parts of the infrastructure (see [](#dos-cp)).
 
 
-## Assumptions
-
-The following assumptions relate to these security considerations. It is assumed that:
-
-1. Every ISD contains at least one core AS.
-2. ISDs are not internally partitioned, so that each AS can reach every other AS within its ISD.
-3. Parent-Child links form a unidirectional hierarchical topology without cyclic relationships.
-4. An AS is described as 'honest' if its private keys are unknown to the attacker and it uses a unique interface identifier for each link. An honest path is one that only traverses honest ASes. A honest segment is the one created by an honest AS.
-
 ## Security Properties {#security-properties}
 
-The properties to be provided by the SCION control plane are:
+
+The SCION control plane provides various security properties, as discussed below. For such properties to hold, ISDs SHOULD NOT be internally partitioned, so that each AS can reach every other AS within its ISD. In addition, an AS is described as 'honest' if its private keys are unknown to the attacker and if it uses a unique interface identifier for each link. An honest path is one that only traverses honest ASes. A honest segment is the one created by an honest AS.
 
 - Connectivity - For every pair of honest ASes X and Y, X will eventually register enough segments to build at least one path (of any length) leading to Y.
 - Forwarding Path Consistency - For every honest path segment registered in any AS
@@ -2069,7 +2061,7 @@ The ISD and SCION AS number are SCION-specific numbers. They are currently alloc
 # Acknowledgments
 {:numbered="false"}
 
-Many thanks go to William Boye (Swiss National Bank), Matthias Frei (SCION Association), Kevin Meynell (SCION Association), Juan A. Garcia Prado (ETH Zurich), and Roger Lapuh (Extreme Networks) for reviewing this document. We also thank Daniel Pascual Galán and Christoph Sprenger from the Information Security Group at ETH Zurich for their inputs based on their formal verification work on SCION. We are also very grateful to Adrian Perrig (ETH Zurich), for providing guidance and feedback about every aspect of SCION. Finally, we are indebted to the SCION development teams of Anapaya, ETH Zurich, and the SCION Association for their practical knowledge and for the documentation about the SCION Control Plane, as well as to the authors of [CHUAT22] - the book is an important source of input and inspiration for this draft.
+Many thanks go to William Boye (Swiss National Bank), Matthias Frei (SCION Association), Kevin Meynell (SCION Association), Juan A. Garcia Prado (ETH Zurich), and Roger Lapuh (Extreme Networks) for reviewing this document. We also thank Daniel Galán Pascual and Christoph Sprenger from the Information Security Group at ETH Zurich for their inputs based on their formal verification work on SCION. We are also very grateful to Adrian Perrig (ETH Zurich), for providing guidance and feedback about every aspect of SCION. Finally, we are indebted to the SCION development teams of Anapaya, ETH Zurich, and the SCION Association for their practical knowledge and for the documentation about the SCION Control Plane, as well as to the authors of [CHUAT22] - the book is an important source of input and inspiration for this draft.
 
 # Deployment Testing: SCIONLab
 {:numbered="false"}
--- a/draft-dekater-scion-controlplane.md
+++ b/draft-dekater-scion-controlplane.md
@@ -1955,12 +1955,21 @@ This section focuses on three kinds of security risks in the control plane:
 2. When "ordinary" (non-core) adversaries try to manipulate the beaconing process (see [](#manipulate-beaconing)).
 3. Denial of Services (DoS) attacks where attackers overload different parts of the infrastructure (see [](#dos-cp)).
 
-These security considerations make assumptions about the formal model of the SCION implementation as follows:
+# Security Assumptions
 
-1. Every ISD has at least one core AS.
+These security considerations make assumptions about the formal model of SCION implementation as follows:
+
+1. Every ISD contains at least one core AS.
 2. The physical links within an ISD are not partitioned and each AS can reach every other AS in the same ISD.
 3. Customer-Provider relationships and their related links form a unidirectional hierarchical topology with cyclic relationships.
 
+An AS is described as 'honest' if its long-term keys are unknown to the attacker and it uses a unique interface identifier for each link. An honest path is one that only traverses honest ASes as follows:
+
+- Connectivity - For every pair of honest ASes X and Y, X will eventually register enough segments to build at least one path (of any length) leading to Y.
+- Path Consistency - For every honest segment registered in any AS, its sequence of AS entries corresponds to a continuous path in the network of physical links and the network topology remains unchanged since the segment was first generated.
+- Loop Freedom - For every honest segment registered in any AS, its sequence of AS entries contains no duplicates, including current and next ISD-AS and interfaces.
+- Beacon Authorization - For every honest segment registered in any AS and any AS X appearing on that segment (except for the previous one), AS X propagated a PCB corresponding to the segment prefix ending in its own entry to its successor AS on the segment.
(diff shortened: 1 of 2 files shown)


draft-dekater-scion-controlplane.md
```diff
@@ -1955,6 +1956,22 @@ This section focuses on three kinds of security risks in the control plane:
 2. When "ordinary" (non-core) adversaries try to manipulate the beaconing process (see [](#manipulate-beaconing)).
 3. Denial of Services (DoS) attacks where attackers overload different parts of the infrastructure (see [](#dos-cp)).
 
+# Security Assumptions
+
+These security considerations make assumptions about the formal model of SCION implementation as follows:
+
+1. Every ISD contains at least one core AS.
+2. The physical links within an ISD are not partitioned and each AS can reach every other AS in the same ISD.
+3. Customer-Provider relationships and their related links form a unidirectional hierarchical topology with cyclic relationships.
```
nicorusti: To be discussed with @dgalan7 (in email thread). I don't think we need this assumption

nicorusti (commented):
I did some rewording, and I'd like to hear back about loop avoidance and on general feedback from @dgalan7 before merging.

draft-dekater-scion-controlplane.md
dgalan7: What a coincidence, I just received this comment's notification as I sent you an email. The email explains why this assumption isn't necessary for loop avoidance - rather, it ensures connectivity by allowing each AS to find a progressively higher-tier ASes until reaching tier 1. We can continue this discussion here or by email.

nicorusti: Absolutely a coincidence, I was writing you an email about this at the very same time, but you were faster. Thank you, your email clarified this. I'll leave the text there (albeit a bit corrected)

nicorusti: I'm still not convinced on this, I will chime in the email thread

draft-dekater-scion-controlplane.md
```diff
@@ -1948,12 +1948,32 @@ Authentication of SCMP packets is not specified here. In current deployments it
...
 2. When "ordinary" (non-core) adversaries try to manipulate the beaconing process (see [](#manipulate-beaconing)).
 3. Denial of Services (DoS) attacks where attackers overload different parts of the infrastructure (see [](#dos-cp)).
 
+
+## Assumptions
+
+The following assumptions relate to these security considerations. It is assumed that:
+
+1. Every ISD contains at least one core AS.
+2. ISDs are not internally partitioned, so that each AS can reach every other AS within its ISD.
```
nicorusti, line 1963: We need to review this assumption. See @tzaeschke 's [comment](https://github.com/scionassociation/scion-cp_I-D/issues/46#issuecomment-2634412040) in another issue

nicorusti: Maybe like this: 
```suggestion
2. ISDs are not partitioned, so that each AS can reach every other AS within its ISD.
```

draft-dekater-scion-controlplane.md
```diff
@@ -1948,12 +1948,32 @@ Authentication of SCMP packets is not specified here. In current deployments it
 
 As described previously, the goal of SCION’s beaconing process in the control plane is to securely discover and disseminate paths between any two ASes. This section describes security considerations for SCION's Control Plane that focuses on *inter*-domain routing. SCION does not provide intra-domain routing, nor does it provide end-to-end payload encryption so these topics lie outside the scope of this section.
 
-This section focuses on three kinds of security risks in the control plane:
+This section discusses three kinds threats to the control plane:
```
tzaeschke, line 1951: kinds _of_ threats

draft-dekater-scion-controlplane.md
```diff
@@ -1948,12 +1948,32 @@ Authentication of SCMP packets is not specified here. In current deployments it
...
+
+- Connectivity - For every pair of honest ASes X and Y, X will eventually register enough segments to build at least one path (of any length) leading to Y.
+- Path Consistency - For every honest segment registered in any AS, its sequence of AS entries corresponds to a continuous path in the network of inter-domain links and the inter-domain network topology remains unchanged since the segment was first generated.
+- Loop Freedom - For every honest segment registered in any AS, its sequence of AS entries contains no duplicates, including current and next ISD-AS and interfaces.
+- Path Authorization - For every honest segment registered in any AS and any AS X appearing on that segment (except for the previous one), AS X propagated a PCB corresponding to the segment portion ending in its own entry to its successor AS on the segment.
+- Path Discoverability - For every directed honest path in the network of inter-domain links, where all traversed ASes permit forwarding there exists at least one protocol execution that registers a segment containing or exactly matching this path, provided the involved links remain stable throughout the protocol execution.
```
tzaeschke, line 1975: I am not sure this is true. PCB forwarding is subject to strict deterministic constraints/limits. I think there are definitely valid _potential_ segments that will never be actualized...? 

Also, I find the terminology slightly confusing. Usually a "path" is the result of stitching up to 3 segments together. Here, it seems that a "path" is something more abstract, i.e. the precursor of one or more segments....?

draft-dekater-scion-controlplane.md
```diff
@@ -1948,12 +1948,32 @@ Authentication of SCMP packets is not specified here. In current deployments it
...
+2. ISDs are not internally partitioned, so that each AS can reach every other AS within its ISD.
+3. Parent-Child links form a unidirectional hierarchical topology without cyclic relationships.
+4. An AS is described as 'honest' if its private keys are unknown to the attacker and it uses a unique interface identifier for each link. An honest path is one that only traverses honest ASes. A honest segment is the one created by an honest AS.
+
+## Security Properties {#security-properties}
+
+The properties to be provided by the SCION control plane are:
+
+- Connectivity - For every pair of honest ASes X and Y, X will eventually register enough segments to build at least one path (of any length) leading to Y.
+- Path Consistency - For every honest segment registered in any AS, its sequence of AS entries corresponds to a continuous path in the network of inter-domain links and the inter-domain network topology remains unchanged since the segment was first generated.
```
tzaeschke, line 1972: I think the sentence is a bit confusing/unclear to me.
I assume there is a split is as follows?

```
For every honest segment registered in any AS:
  - its sequence of AS entries corresponds to a continuous path in the network of inter-domain links
  - the inter-domain network topology remains unchanged since the segment was first generated.
```
Maybe insert a comma or semicolon after "inter-domain links"?

draft-dekater-scion-controlplane.md
tzaeschke: Just out of curiosity, so the plan is to disallow partitioned ISDs?

draft-dekater-scion-controlplane.md
nicorusti: Adopted your suggestion

draft-dekater-scion-controlplane.md
tzaeschke: Apologies if I don't quite understand. If the assumptions do not match the actual specified system, why are we discussing them in the RFC?
- Are they indicative of future changes to the spec?
- Are they recommendations for users? E.g. "if you want a certain security property, do not partition your ISD"?

It may be worth clarifying this to any readers, explicitly stating that the assumptions do not apply to the system (unless you put additional restrictions on the system such as not partitioning your ISD). Otherwise readers may get really confused...? 

Also, is it a valid attack vector if someone somehow moves their AS outside of the ISD, effectively creating a partition?

Again, apologies if I miss the point here...

draft-dekater-scion-controlplane.md
tzaeschke: Yes, I tried to attach the comment to "path discoverability" apparently that didn't work.
>  I think the keyword here is _where all traversed ASes permit forwarding _

The problem is that ASes forward only 5 segments per round. Yes, they may be multiple rounds, but at some point the segments expire. Once segments expire, they should forward them again (assuming they are superior to other segments which may have not been forwarded yet). With an expiration time of, say, 1 hour and forwarding every 60 seconds, we get a maximum of 60*5 = 300 segments. If we have more than 300 segments, then we will never forward all of them. 

If we permitted forwarding _all_ segments, that may overload the network (the number of paths growths exponentially). So we cannot physically forward all segments because that would be an attack in itself.

If "permit" is more liberal here then previously specified, maybe rephrase to "where all traversed ASes permit _unlimited_ forwarding"?
Otherwise I think a reader would interpreted "permit" as specified in the section "Selection of PCBs to Propagate".
Again, similar to the point I made earlier, I am not sure why we list assumptions here that, as I understand, do not apply to the specified system. It may help to have an more explicit statement that explains that some assumptions do not apply and why we list them anyway...?
I feel I am missing something important here.

> I updated wording to follow the terminology in the draft ("path segment", "forwarding path"). Is it better?
Thanks, yes :)

tzaeschke: Sorry, I just realized that this is a "property" not an "assumption". I think my argument still holds though.

draft-dekater-scion-controlplane.md
tzaeschke: I just talked to the team here, and apparently consensus is that ISD should indeed be disallowed.
If that is the case, then the assumption would be correct, but the spec may need updating, I don't think it says anywhere that ISD partitions are not allowed?

nicorusti: > Apologies if I don't quite understand. If the assumptions do not match the actual specified system, why are we discussing them in the RFC?

 Because, based on the review we received by the formal methods group, they clearly recommend clarify assumptions and security properties, I forwarded you what we received. 
 
 How about we reword the assumptions section in general?  We now have: 
 ```
The following assumptions relate to these security considerations. It is assumed that:

1. Every ISD contains at least one core AS.
2. ISDs are not internally partitioned, so that each AS can reach every other AS within its ISD.
3. Parent-Child links form a unidirectional hierarchical topology without cyclic relationships.
4. An AS is described as 'honest' if its private keys are unknown to the attacker and it uses a unique interface identifier for each link. An honest path is one that only traverses honest ASes. A honest segment is the one created by an honest AS.
```

I think for these assumptions: 

1. We don't need it as we already say in the draft that an ISD MUST contain a core AS. 
2. We could reword it, saying that for the security properties to hold, an ISD SHOULD not be internally partitioned. 
3. I don't think we need this one. The beaconing algorithm stops re-propagating beacons if there is a loop, handling this and making the topology acyclic. Therefore, every AS then finds its way to the core through the acyclic graph.  I tried hard to think of an example where a loop would isolate an AS, and I couldn’t. (Unless I’m missing something)
4. THis is not really an assumption, rather a definition of how we sue "honest"

Based on my thought above, we could rewrite the assumption paragraph like this: 

> For the following security properties to hold, ISDs should not be internally partitioned, so that each AS can reach every other AS within its ISD. In addition,  an AS is described as 'honest' if its private keys are unknown to the attacker and if it uses a unique interface identifier for each link. An honest path is one that only traverses honest ASes. A honest segment is the one created by an honest AS.

@tzaeschke What do you think? THis way we are also not too prescriptive about wether an ISD should be partitioned or not.

tzaeschke: Yes, I think that sounds good.
I have one more nitpick though: I assume for the security properties to hold, ISDs **must not** (or even MUST NOT) be internally partitioned? Usually assumptions either must hold or they don't matter?
Maybe @dgalan7 can comment on this?

dgalan7: From my email:

> Some time ago we had a technical discussion with Adrian Perrig, and we agreed that real SCION deployments would probably want to disallow partitioned ISDs altogether, regardless of our model and properties. We have noted this assumption to remind you in case it is missing. If however the consensus is that partitioned ISDs are to be allowed in the real world, it looks sensible to choose a "should not" (or perhaps even a "SHOULD NOT") instead and explicitly link it to the properties, as you suggest. Specifically, proving the Connectivity property needs this assumption.

draft-dekater-scion-controlplane.md
dgalan7: From my email:

> We are aware of the limits, including the quantity limit in beacon stores, the quantity limit on selection sizes (best PCB set size), and the time limits due to fixed propagation/registration periods. So far, none of these limits are present in our model, and as Tilmann (@tzaeschke) notes, the properties would hold when considering an unlimited context with no limits. We plan to model limits as future work, especially since we agree with Adrian Perrig that they can become attack vectors. For this Internet-Draft, I guess the best that can be done is to either fixing specific numbers as limit recommendations (like you are already doing) or to abstractly say that limits should not be unreasonably strict.

draft-dekater-scion-controlplane.md
nicorusti: @dgalan7 Great, I updated the text to use SHOULD NOT.

> I have one more nitpick though: I assume for the security properties to hold, ISDs must not (or even MUST NOT) be internally partitioned? Usually assumptions either must hold or they don't matter?

@tzaeschke to address your concern, how about we add some wording on the impact of partitioning? I'm thinking something like: 
"For such properties to hold, ISDs SHOULD NOT be internally partitioned, so that each AS can reach every other AS within its ISD. **In case an AS gets partitioned from its ISD, if it is non-core, then it won't be able to reach other ASes within the ISD. If it is core, it may reach other ASes while transiting via other ISDs if permitted by its policies. In such case, transiting via another ISD may potentially allow external agents to influence the internal connectivity of an ISD.** 

This was a quick and dirty wording, I'd need to think a bit more about it

draft-dekater-scion-controlplane.md
dgalan7: The danger of removing this property is that there could be a mistake in the protocol's description that makes it impossible to discover some links under some situations (even when limits and policies would otherwise allow it), and there would be nothing in this I-D stating that it is unintended behavior. Depending on how acceptable that is, it could be removed or rephrased to something that is weaker, but hints that in principle the protocol should not "miss" links with the limits or the policies explicitly causing this behavior.

draft-dekater-scion-controlplane.md
nicorusti: As discussed in the email, 
> The assumption that business relationships form an acyclic graph is not needed. 

Resolving as I removed assumption 3 already.

draft-dekater-scion-controlplane.md
nicorusti: From the email thread, from Daniel:

> Furthermore, we suggest removing the previous assumption that every ISD should not be internally partitioned. We replace both with this:
>
> To guarantee that every ISD is internally connected (each AS reaches all other ISD members):
1a. Cores of  the ISD can reach each other without leaving the ISD, i.e., for every pair of cores in an ISD there is a physical sequence of links that only traverses ISD members.
> 2. For every non-core AS in the ISD, there is at least one sequence of customer-to-provider links going "up" to some core.
>
>To guarantee connectivity in all the world (all the considered network):
1b. All cores can reach each other with some physical sequence of links.
> 2. (Same as before)
>
> We consider several combinations that may appear in the Internet-Draft and their combinations:
> - If (1a) and (2) are required, then every ISD will be internally connected, but it will not be possible for ISDs to be "partitioned" in the sense that some path escapes the ISD and then re-enters it later. We understand that this is too strict for current SCION.
> - If all (1a), (1b) and (2) are required, then all the network will be connected, every ISD will be internally connected, and it will be forbidden to have paths escaping ISDs and re-entering them later. This is also too strict for current SCION.
> - If (1b) and (2) are required, then all the network will be connected, but there will be no guarantee of internal connectivity for ISDs. Consequently, ISD members will be allowed to (and may need to) use paths that escape their ISD and re-enter it later. To the best of our understanding, this seems to be what aligns the best with the intended control plane behavior.
> - Having (1a) or (1b) without (2) seems unlikely to provide any guarantees, so we do not advise these combinations.
>
> A possible way of framing this in the Internet-Draft could be to say something similar to "To ensure that the properties hold in all the network, it should be the case that (1b) and (2)". Optionally, "Furthermore, to ensure the properties in the context of a single ISD, it should also be the case that (1a)". As discussed in the PRs, this is possibly best to be left as recommendations that informatively guide concerned stakeholders into voluntary steps that lead to stronger notions of correctness and security."

nicorusti: I added this wording, based on what @dgalan7 suggested: 

> To ensure that the properties hold across the overall SCION network, all core ASes should be able to reach each other with some sequence of core links, and all non-core ASes should have at least one path up to a core AS. Furthermore, to ensure that the properties hold within a single ISD, all cores ASes of the ISD should be able to reach each other without leaving the ISD, i.e., for every pair of cores in an ISD there is a sequence of SCION links that only traverses ISD members.
A core AS may reach other core ASes in the same ISD via other ISDs. This may be permitted, depending on the ISD's policies.  

@dgalan7 @tzaeschke what do you think? I hope we are converging :)

draft-dekater-scion-controlplane.md
nicorusti: > The risk of removing this property is that there could be a mistake in the protocol's description that makes it impossible to discover some links under some situations 

in the draft we already recommend limits in section "2.3.1.1.  Storing and Selecting Candidate PCBs”.  I think the best approach is to not have this property in the security considerations, and rather add some wording on the tradeoffs in 2.3.1.1, when we talk about PCB set size.

nicorusti: @dgalan7 @tzaeschke how do you feel about the current wording I added: 

> Note that the PCBs set size should not be too low, in order to make sure that beaconing can discover a wide amount of paths. Values above are RECOMMENDED maxima which represent a tradeoff between scalability and amount of paths discovered. In current practice the intra-ISD set size is typically 20.

tzaeschke: LGTM

nicorusti: I will now merge, as we are approaching the deadline. We can touch on this later on if needed.

nicorusti (approved):

PR #98, opened 2025-02-25
Clarify beaconing ckecks

### LINKED ISSUE #90: Beaconing: clarify checks
- [x] Check if these checks are implemented @jiceatscion 
- [x] Do changes proposed in draft
- [x] Clarify if last check (remote registration) is actually needed

Work on checks at beaconing was previously done in #18 

___ 
Feedback David Galan: 
3.  We propose that the local checks performed by each AS, necessary to ensure protocol correctness, are clarified. Currently, the draft requires that honest ASes validate the signatures in PCBs. However, honest ASes may accept malformed PCBs, even after a successful signature validation, unless they also perform some additional local checks. These local checks occur mainly in two key moments of the protocol:

- (Reception) Subsection 2.3.2.1 "Reception of PCBs" details the local checks performed when an AS received a beacon propagated by another AS. We have synthesized a minimal set of additional checks that are required for protocol correctness, and we propose their inclusion in the I-D:

    - (Cores) During core beaconing, the receiver Control Service must check every AS entry and discard PCBs that were propagated at any point (according to the `signed.header_and_body.body.isd_as' field) by a non-core AS.
    - (Interfaces) If the received PCB was (a) last propagated by AS X (according to the `signed.header_and_body.body.isd_as' field of the last AS entry), and (b) was received through the local AS interface i (according to local knowledge of the Control Service), then the local knowledge of the receiver Control Service MUST agree that the link behind interface i indeed leads to AS X.
    - (Continuity) When a PCB contains two or more AS entries, the receiver Control Service must check every AS entry except the last and discard beacons where the field `signed.header_and_body.body.next_isd_as' of an entry does not equal the field `signed.header_and_body.body.isd_as' of the next entry

- (Remote registration) In subsection 4.1 "Intra-ISD Path Segment Registration" consider mentioning that, during remote registration (a phase of intra-ISD beaconing) of a PCB from an AS X into the database of an AS Y, AS Y must not accept the PCB if an AS other than Y generated it (field `signed.header_and_body.body.isd_as' of the first AS entry).

nicorusti:
FIxes #90

I'm not decided on wether the check on registration and the first check at reception is necessary, as discussed in the issue. @jiceatscion can you help me sort it out? 

And @knmeynell can you check the language a bit? Some concepts were not easy to express, and I tried to simplify David's text

### DIFF
--- a/draft-dekater-scion-controlplane.md
+++ b/draft-dekater-scion-controlplane.md
@@ -1423,7 +1423,7 @@ Every registration period, the Control Service of a non-core AS performs the fol
 
 1. The Control Service selects the PCBs that it wants to transform into down segments from the candidate PCBs in the Beacon Store.
 2. The Control Service "terminates" the selected PCBs by performing the steps described in [](#term-pcb). From this moment on, the modified PCBs are called **down segments**.
-3. The Control Service registers the newly created down segments with the Control Services of the core ASes that originated the corresponding PCBs. This is done by invoking the `SegmentRegistrationService.SegmentsRegistration` remote procedure call (RPC) in the Control Services of the relevant core ASes (see also [](#reg-proto)). The first ISD-AS entry of the path segment MUST be equal to the core ISD-AS where the segment is being registered. If not, the core AS MUST reject the segment.
+3. The Control Service registers the newly created down segments with the Control Services of the core ASes that originated the corresponding PCBs. This is done by invoking the `SegmentRegistrationService.SegmentsRegistration` remote procedure call (RPC) in the Control Services of the relevant core ASes (see also [](#reg-proto)). The first ISD-AS entry of the path segment SHOULD be equal to the core ISD-AS where the segment is being registered. If not, the core AS SHOULD reject the segment.
 
 **Note:** For more information on possible selection strategies of PCBs, see [](#selection).
 
--- a/draft-dekater-scion-controlplane.md
+++ b/draft-dekater-scion-controlplane.md
@@ -1218,11 +1218,14 @@ To bootstrap the initial communication with a neighboring beacon service, ASes u
 
 #### Reception of PCBs
 
-The following first steps of the propagation procedure are the same for both intra-ISD and core beaconing:
+Upon receiving a PCB, the Control Service of an AS performs the following checks:
 
-1. Upon receiving a PCB, the Control Service of an AS verifies the validity of the PCB (see [](#pcb-validity)) and invalid PCBs MUST be discarded. The PCB contains the version numbers of the TRC(s) and certificate(s) that MUST be used to verify its signatures which enables the Control Service to check whether it has the relevant TRC(s) and certificate(s). If not, they can be requested from the Control Service of the sending AS.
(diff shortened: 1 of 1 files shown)


jiceatscion (approved):

draft-dekater-scion-controlplane.md
```diff
@@ -1420,7 +1423,7 @@ Every registration period, the Control Service of a non-core AS performs the fol
 
 1. The Control Service selects the PCBs that it wants to transform into down segments from the candidate PCBs in the Beacon Store.
 2. The Control Service "terminates" the selected PCBs by performing the steps described in [](#term-pcb). From this moment on, the modified PCBs are called **down segments**.
-3. The Control Service registers the newly created down segments with the Control Services of the core ASes that originated the corresponding PCBs. This is done by invoking the `SegmentRegistrationService.SegmentsRegistration` remote procedure call (RPC) in the Control Services of the relevant core ASes (see also [](#reg-proto)).
+3. The Control Service registers the newly created down segments with the Control Services of the core ASes that originated the corresponding PCBs. This is done by invoking the `SegmentRegistrationService.SegmentsRegistration` remote procedure call (RPC) in the Control Services of the relevant core ASes (see also [](#reg-proto)). The first ISD-AS entry of the path segment MUST be equal to the core ISD-AS where the segment is being registered. If not, the core AS MUST reject the segment.
```
nicorusti, line 1426: As being discussed in #90, @jiceatscion and I are not sure this check is necessary. As JC writes, 
> After all the whole segment's validity can be verified independently from how it was obtained, is it not? How does the fact that a segment is only served by its legitimate originator add any guarantee?

So this control does not add much. Maybe it could be changed to SHOULD?

nicorusti, line 1426: As being discussed in #90, @jiceatscion and I are not sure this check is necessary. As JC writes, 
> After all the whole segment's validity can be verified independently from how it was obtained, is it not? How does the fact that a segment is only served by its legitimate originator add any guarantee?

So this control does not add much. Maybe it could be changed to SHOULD?

```suggestion
3. The Control Service registers the newly created down segments with the Control Services of the core ASes that originated the corresponding PCBs. This is done by invoking the `SegmentRegistrationService.SegmentsRegistration` remote procedure call (RPC) in the Control Services of the relevant core ASes (see also [](#reg-proto)). The first ISD-AS entry of the path segment SHOULD be equal to the core ISD-AS where the segment is being registered. If not, the core AS MUST reject the segment.
```

jiceatscion: "If not, the core AS MUST reject the segment"  -> "If not, the core AS SHOULD reject the segment. For consistency?

jiceatscion: Other than that. LGTM.

knmeynell: The language looks pretty good to me and it's understandable.

nicorusti: Thanks. I asked for an additional opinion on [slack](https://scionproto.slack.com/archives/C8ADA9CEP/p1740679917948679).

nicorusti: No answer there, I guess I'll leave it as it is..

PR #101, opened 2025-02-28
PCB policies: add wording by Daniel Galan

### LINKED ISSUE #16: Beaconing: Clarify PCB selection policies and how they meet endpoint's needs
> From:	Joel Halpern
> There seems to be a philosophical contradiction in section 2.3.1 (Selection of PCBs to propagate) of the control plane document. Given the philosophy of SCION, the goal should be to propagate PCBs with the highest probability of collectively meeting the needs of the end hosts that will perform path construction.  Of course, this AS doesn't know that.   But the text ignores this distinction.

Tis also relates to [rfc9217 section-2.8](https://www.rfc-editor.org/rfc/rfc9217.html#section-2.8):
```how can the incentives of network operators and end users be aligned to realize the vision of path-aware networking, ...```

### PR DESCRIPTION
**Pull request by nicorusti**
Resolves #16

### DIFF
--- a/draft-dekater-scion-controlplane.md
+++ b/draft-dekater-scion-controlplane.md
@@ -1106,6 +1106,7 @@ For the purpose of constructing and propagating path segments, an AS Control Ser
 - Neighbor interface underlay address
 
 In addition, the maximum MTU supported by all intra-AS links MAY be configured.
+Last, the AS SHOULD adopt a PCB selection policy that it does not accidentally isolate the AS from the network, i.e., such that it does not block connectivity to parent providers and that ensures downstream connectivity for children. For more details, see [](#selection-policy-example).
 
 ## Propagation of PCBs {#path-prop}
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: draft-dekater-scion-controlplane.md [hunk 1]
```diff
@@ -1145,9 +1146,10 @@ Note that to ensure quick connectivity establishment, an AS MAY attempt to forwa
 
 The scalability implications of such parameters are further discussed in [](#scalability).
 
-#### Selection Policy Example
+#### Selection Policy Example {#selection-policy-example}
 
-An AS MUST select the best PCBs set to be further propagated. Selection may be based on criteria such as:
+An AS MUST select the best PCBs set to be further propagated.
+A PCB Selection Policy can be expressed as a stateful filter of paths, i.e., a function which indicates whether to accept or deny a given path. This filter is stateful in that it can be updated each time its AS registers a new segment. Selection may be based on criteria such as:
```
jiceatscion, line 1152: May be replace "path[s]" with "segment[s]"?
Other than that, LGTM.

nicorusti:
```suggestion
A PCB Selection Policy can be expressed as a stateful filter of segments, i.e., a function which indicates whether to accept or deny a given path. This filter is stateful in that it can be updated each time its AS registers a new segment. Selection may be based on criteria such as:
```

jiceatscion (approved):

PR #123, opened 2025-06-30
Updated SCION AS table

### LINKED ISSUE #112: Numbers registry - update reference and clarifications
- [x] create placeholder link on scion.org about registry activity (for now redirecting to Anapaya AND linking the same table as the draft (which I think is now on scionproto github)
- [x] Update reference to point to scion.org
- [x] Check reference to RFC4893 in relation to AS number definition
- [x] Do we need a reference to RFC5952 in relation to formatting? Maybe not as Alvaro suggests
- [x] Check that ISD-AS in the drafts follow the text representation rule (e.g. `4-0000:1:f` does not)  
- [x] Move table 2 to 1.5.2.
- [x] https://github.com/scionassociation/scion-cppki_I-D/pull/68 from the PKI draft and replace it with cross  reference: `Table 1: ISD Number Allocations`
- [x] Devs: confirm that the right notation for ISD 65551 is `0:1:f` (which is what we have in the draft)
- [x] Update PKI draft https://github.com/scionassociation/scion-cppki_I-D/pull/73

---- 
### 1.5.1.  ISD Numbers
```
542   An ISD number is the 16-bit global identifier for an ISD and MUST be
543   globally unique.

545   The following table gives an overview of the ISD number allocation:

547     +==============+================================================+
548     | ISD          | Description                                    |
549     +==============+================================================+
550     | 0            | The wildcard ISD.                              |
551     +--------------+------------------------------------------------+
552     | 1 - 15       | Reserved for documentation and sample code     |
553     |              | (analogous to [RFC5398]).                      |
554     +--------------+------------------------------------------------+
555     | 16 - 63      | Private use (analogous to [RFC6996]).  Can be  |
556     |              | used for testing and private deployments       |
557     +--------------+------------------------------------------------+
558     | 64 - 4094    | Public ISDs.  Should be allocated in ascending |
559     |              | order, without gaps and "vanity" numbers.      |
560     +--------------+------------------------------------------------+
561     | 4095 - 65535 | Reserved for future use.                       |
562     +--------------+------------------------------------------------+

564                      Table 1: ISD number allocations

566   Currently, ISD numbers are allocated by Anapaya, a provider of SCION-
567   based networking software and solutions (see [ISD-AS-assignments]).
```

>  Suggestion: the Anapaya site doesn't have a summary similar to Table 1.
If anyone is looking there, they won't know which ISD can be used for
testing (for example).  You might want to include this information, similar
to the IANA pages. ;-)

### 1.5.2.  SCION AS Numbers
```
571   A SCION AS number is the 48-bit identifier for an AS.  Although they
572   play a similar role, there is no relationship between SCION AS
573   numbers and BGP ASNs as defined by [RFC4893].  For historical reasons
574   some SCION Autonomous Systems use a SCION AS number where the first
575   16 bits are 0 and the remaining 32 bits are identical to their BGP
576   ASN, but there is no technical requirement for this.
```
> [] s/RFC4893/RFC4271 I know that there isn't a specific definition in rfc4271 of an ASN (there
isn't one in rfc4893 either).

###  1.5.3.2.  AS numbers
```
587   The text representation of SCION AS numbers is similar to IPv6 (see
588   [RFC5952]) but not identical.  It MUST be as follows:

590   *  It uses a 16-bit colon-separated lower-case hex encoding with
591      leading 0s omitted: 0:0:0 to ffff:ffff:ffff.

593   *  The :: zero-compression feature of IPv6 MUST NOT be used.  The
594      feature has very limited use in a 48-bit address space and would
595      only add more complexity.

597   *  A range of AS numbers can be shortened with a notation similar to
598      the one used for CIDR IP ranges ([RFC4632]).  For example, the
599      range of the lowest 32-bit AS numbers (0-4294967295) can be
600      represented as 0:0:0/16.

602   *  For historical reasons, SCION AS numbers in the lower 32-bit range
603      MAY also be represented as decimal for human readability.  For
604      example, if a program receives the AS number 0:1:f, it MAY display
605      the number as "65551".
```
> IMHO, the comparison to RFC5952 makes the explanation more complex as
you have to then explain the exceptions (::)...

### Example
```
614   For example, the text representation of AS number 65551 (0x1000f) in
615   ISD number 4 is 4-0000:1:f.
```

> 0000:1:f doesn't follow the rules in §1.5.3.2: the leading 0s are
included.  Shouldn't it be 1:f?

### Table 2: AS number allocations

> Maybe this table belongs in §1.5.2.

### LINKED ISSUE #120: Are the ISD and SCION AS number allocations correct?
Section 1.5.1 - ISD numbers

Should the ranges be 64-4095 and 4096-65535?

The first 12-bits provide 4096 numbers, which if starting from zero means the last number in the range should be 4095 rather than 4094. This would mean the range above should start at 4096. 

Section 1.5.5 - Special-Purpose SCION AS Numbers

This section is confused. I think it may already have been moved into Section 1.5.2 "SCION AS numbers" in a pending PR, but I think this should simply be called 'SCION AS number allocations'. 

The ranges of Public AS numbers are not special purpose. 

The allocations have large gaps in the address space - assume these are reserved for future use, but should they be documented as with the reserved ISD numbers? See https://docs.google.com/spreadsheets/d/1p-Qdon6umxY6mVaoPU062NvSxkoeFVW_ulM2ZvwtVyk/

knmeynell:
Added full address ranges and reserved space

Resolves #120 
Resolves #112 (mention AS number formatting MUST use decimal for lower ranges) 

[Diff with main](https://author-tools.ietf.org/api/iddiff?url_1=https://scionassociation.github.io/scion-cp_I-D/draft-dekater-scion-controlplane.txt&url_2=https://scionassociation.github.io/scion-cp_I-D/SCION_AS_Table/draft-dekater-scion-controlplane.txt)


draft-dekater-scion-controlplane.md
```diff
@@ -379,8 +379,7 @@ The text representation of SCION ISD numbers MUST be its decimal ASCII represent
 The text representation of SCION AS numbers MUST be as follows:
 
 - Using big-endian hexadecimal notation in 3 groups of 4, in the range `0:0:0` to `ffff:ffff:ffff`.
-- Leading zeros in each group are omitted (e.g., `0:0:1`)
-- The `::` zero-compression feature of IPv6 MUST NOT be used.
+- Leading zeros in each group are omitted, with the exception that one zero must be notated if a group is entirely zeros (e.g., `0:0:1`). The `::` zero-compression feature of IPv6 MUST NOT be used.
```
nicorusti, line 382:
```suggestion
- Leading zeros in each group are omitted, with the exception that one zero MUST be notated if a group is entirely zeros (e.g., `0:0:1`). The `::` zero-compression feature of IPv6 MUST NOT be used.
```

draft-dekater-scion-controlplane.md
```diff
@@ -358,16 +358,16 @@ SCION endpoints use wildcard AS `0:0:0` to designate any core AS, e.g. to place
...
-| `2:0:0/16`       | ~4.3&nbsp;bill.  | Additional public SCION AS numbers                                     |
-| `ff00:0:0/32`    | 65535       | Reserved for documentation and test/sample code (analogous to {{RFC5398}}). |
-| `ffaa:0:0/24`    | ~16.8&nbsp;mill. | Reserved for private use (analogous to {{RFC6996}}). These numbers can be used for testing/private deployments. |
+| `0:0:1 - 0:ffff:ffff`| ~4.3&nbsp;billion| Public SCION AS numbers                                              |
+| `1:0:0 - 1:ffff:ffff`| ~4.3&nbsp;billion| Reserved for future use                                              |
+| `2:0:0 - 2:ffff:ffff`| ~4.3&nbsp;billion| Public SCION AS numbers                                              |
+| `3:0:0 - feff:ffff:ffff`| ~280&nbsp;trillion| Reserved for future use                                              |
+| `ff00:0:0 - ff00:0:ffff`| 65536      | Reserved for documentation and sample code (analogous to {{RFC5398}}). |
+| `ffaa:0:0 - ffaa:ff:ffff`| ~16.8&nbsp;million| Reserved for private use (analogous to {{RFC6996}}). These numbers can be used for testing/private deployments. |
```
nicorusti, line 366:
```suggestion
| `ff00:0:0 - ff00:0:ffff`| 65536      | Reserved for documentation and sample code (analogous to {{RFC5398}}). |
| `ff00:1:0 - ffa9:ffff:ffff`| ~281&nbsp;trillion | Reserved |
| `ffaa:0:0 - ffaa:ff:ffff`| ~16.8&nbsp;million| Reserved for private use (analogous to {{RFC6996}}). These numbers can be used for testing/private deployments. |
```

I think you are missing one range here

draft-dekater-scion-controlplane.md
```diff
@@ -358,16 +358,16 @@ SCION endpoints use wildcard AS `0:0:0` to designate any core AS, e.g. to place
...
+| `ffaa:100:0 - ffff:ffff:fffe`| ~369&nbsp;billion| Reserved for future use                                     |
 | `ffff:ffff:ffff` | 1           | Reserved                                                                    |
```
nicorusti, line 368: What is the difference between "reserved for future use" and Reserved (e.g. `ffff:ffff:ffff`)?
Shall we use "unallocated" instead of reserved? Because these spaces are actually unallocated, they are not truly reserved for anything..

nicorusti (changes requested):
See comments

draft-dekater-scion-controlplane.md
knmeynell: 'Reserved for future use' was the terminology used in the ISD table. I've changed in both to 'Unallocated'.

draft-dekater-scion-controlplane.md
knmeynell: I've added this.

knmeynell (commented):
Fine

jiceatscion (approved):
Except for my opinion about the decimal representation which I posted in the issue, LGTM.

knmeynell:
> Except for my opinion about the decimal representation which I posted in the issue, LGTM.

I agree use of MUST and MAY is somewhat contradictory. I've revised this a bit.

draft-dekater-scion-controlplane.md
```diff
@@ -381,13 +382,13 @@ The text representation of SCION ISD numbers MUST be its decimal ASCII represent
 
 #### AS numbers
 
-The text representation of SCION AS numbers MUST be as follows:
+The text representation of SCION AS numbers SHOULD be as follows:
+
+- SCION AS numbers in the lower 32-bit range SHOULD be represented as decimal notation.
+- SCION AS numbers in the higher 32-bit range MUST be represented using big-endian hexadecimal notation in 3 groups of 4, in the range `1:0:0` to `ffff:ffff:ffff`. Leading zeros in each group are omitted, with the exception that one zero MUST be notated if a group is entirely zeros (e.g., `1:0:1`). The `::` zero-compression feature of IPv6 MUST NOT be used.
+- Implementations MAY accept AS numbers in the lower 32-bit range in hexadecimal notation, but they should convert them to decimal notation (e.g. a program may accept AS number '0:1:f' and convert it to 65551).
+- A range of AS numbers can be shortened with a notation similar to the one used for CIDR IP ranges ({{RFC4632}}). In such case, hexadecimal notation MUST be used. For example, the range of the lowest 32-bit AS numbers (0-4294967295) can be represented as `0:0:0/16`.
```
nicorusti, line 390:
```suggestion
The text representation of SCION AS numbers is as follows:

- SCION AS numbers in the lower 32-bit range MUST be represented as decimal notation when output by implementations. Implementations MAY accept AS numbers in the lower 32-bit range in hexadecimal notation too, but they should convert them to decimal notation (e.g., a program may accept AS number '0:1:f' and convert it to 65551).
- SCION AS numbers in the higher 32-bit range MUST be represented using big-endian hexadecimal notation in 3 groups of 4, in the range `1:0:0` to `ffff:ffff:ffff`. Leading zeros in each group are omitted, with the exception that one zero MUST be notated if a group is entirely zeros (e.g., `1:0:1`). The `::` zero-compression feature of IPv6 MUST NOT be used.
- A range of AS numbers can be shortened with a notation similar to the one used for CIDR IP ranges ({{RFC4632}}). In such case, hexadecimal notation MUST be used. For example, the range of the lowest 32-bit AS numbers (0-4294967295) can be represented as `0:0:0/16`.
```

nicorusti: @jiceatscion @knmeynell what do you think about this wording? 
- I removed the first MUST
- I tried to use the woridng suggested by @jiceatscion

jiceatscion: I still disagree with the following phrase:
"but they should convert them to decimal notation (e.g., a program may accept AS number '0:1:f' and convert it to 65551)."

The problem I see with this is that it does not state at all what the conversion is for and when it may take place. Internally neither string representations are used. Any conversion that takes place is from string to internal (when parsing config or args from humans) and from internal to string (when printing out human readable information or generating config). So, conversions between the two string representations aren't relevant. We must state what is printed and what is parsed. That's it.

jiceatscion: LGTM

draft-dekater-scion-controlplane.md
```diff
@@ -394,7 +394,7 @@ The text representation of SCION AS numbers SHOULD be as follows:
 
 The text representation of SCION addresses MUST be `<ISD>-<AS>`, where `<ISD>` is the text representation of the ISD number, `<AS>` is the text representation of the AS number, and `-` is the literal ASCII character 0x2D.
 
-For example, the text representation of AS number 65551 (0x1000f) in ISD number 4 is `4-0:1:f`.
+For example, the text representation of AS number 280375465082881 (0xff0000000001) in ISD number 15 is `15-ff00:0:1`.
```
nicorusti, line 397: We should use the text representation according to rules above, so higher-range AS numbers in the documentation range MUST be printed in HEX. I replaced it again with `ff00:0:1`

nicorusti (approved):

PR #130, opened 2025-07-07
Remove external SVGs and use aasvg for SCMP diagrams

nicorusti:


nicorusti:
LGTM, TXT version is the same

Screenshot 2025-07-07 at 20 02 22

And aasvg SCMP does not look too bad..

Screenshot 2025-07-07 at 20 02 51

PR #145, opened 2025-10-15
Remove duplicate gRPC definitions

### LINKED ISSUE #67: protobuf RPC API - remove duplicated definitions in appendix
With #24 we introduced a description of the gRPC CP API in the appendix.  However, with this, some definitions are duplicated. For example, `ASEntrySignedBody` is defined in both `2.2.1.4.2.  AS Entry Signed Body` and in the appendix. 
 
TODO: 
- [x] evaluate wether we leave the gRPC API descriptions all in the appendix or in various paragraphs --> we leave them in both places
- [x] ~some gRPC definitions are duplicated, duplicates should be removed (preferably, leave them in the respective chapters rather than in appendix). A non-exhaustive list:~
  - `SegmentType`
  - `SegmentsRegistrationRequest`
  - `SegmentCreationService`
  - `BeaconRequest`
  - `BeaconResponse`
  - `PathSegment`
  - `SegmentInformation`
  - `ASEntry`
  - `ASEntrySignedBody`
  - `HopEntry`
  - `PeerEntry`
  - `HopField`
  - `SignedMessage`
  - `Header`
  - `HeaderAndBodyInternal`
  - `VerificationKeyID`
- [x] make sure there are proper cross-references to the respective gRPC requests For example: 
  - `SegmentsRequest` right now it is not clear which requests are used for path lookups by endpoints, e.g. we should mention that an endpoint sends a `SegmentsRequest` to its path server.
  - `SegmentsResponse`
  - `SignatureAlgorithm`
- [x] We sometimes call the API descritpions in the text "on Code-Level", we should use a better wording (e.g. gRPC API)
- [x] check descriptions and typos  (e.g. `// This API is exposed by the control services of core ASes expose this on the SCION dataplane and also by all`

### LINKED ISSUE #96: The PKI gRPC API should be specified
We do not specify how certificates are requested between ASes. Right now we reference the PKI draft, but also there we don't say anything. We should add this somehwere (e.g. together with the gRPC API)

Pointers:
- Documentation https://docs.scion.org/en/latest/cryptography/interactions.html#messages 
- Implementation: https://github.com/scionproto/scion/blob/abd39afe0a83199aca5ad59d6cc4cc383c89d4cf/proto/control_plane/v1/cppki.proto#L55

knmeynell:
Resolves #67 
Resolves #96 
[Diff with Main](https://author-tools.ietf.org/api/iddiff?url_1=https://scionassociation.github.io/scion-cp_I-D/draft-dekater-scion-controlplane.txt&url_2=https://scionassociation.github.io/scion-cp_I-D/Removed-and-rationalised-gRPC-references/draft-dekater-scion-controlplane.txt)

### DIFF
--- a/draft-dekater-scion-controlplane.md
+++ b/draft-dekater-scion-controlplane.md
@@ -1331,27 +1331,20 @@ message Chain {
 }
 ~~~~~
 
-A `ChainsRequest` message includes following fields:
+A `ChainsRequest` message includes the following fields:
 
 - `isd_as`: Returns ISD-AS of Subject in the AS certificate.
 - `subject_key_id`: Returns SubjectKeyID in the AS certificate.
 - `at_least_valid_until`: Point in time at which the AS certificate must still be valid - in seconds since UNIX epoch.
 - `at_least_valid_since`: Point in time at which the AS certificate must be or must have been valid - in seconds since UNIX epoch.
 
-A `ChainsResponse` includes following fields:
+A `ChainsResponse` includes the following fields:
 
 - `chains`: Lists the certificate chains that match the request. A `Chain` contains:
   - `as_cert`: Returns the AS certificate in the chain.
   - `ca_cert`: Returns the CA certificate in the chain.
 
-
-A `TRCRequest` includes following fields:
-
-- `isd`: Returns the ISD number of the TRC.
-- `base`: Returns the base number of the TRC.
-- `serial`: Returns the serial number of the TRC.
-
-The returned `trc` contains the raw TRC.
+For requesting TRCs, the protobuf messages are:
 
 ~~~~~
 message TRCRequest {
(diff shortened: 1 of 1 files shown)


nicorusti (commented):
I reordered fields a bit, what do you think? I want to give it another look. 

Also, can you update the reference in the PKI ID?

draft-dekater-scion-controlplane.md
```diff
@@ -1588,7 +1661,36 @@ The overall sequence of requests to resolve a path SHOULD be as follows:
...
+
+    // Mapping from path segment type to path segments.
+    // The key is the integer representation of the SegmentType enum.
+    map<int32, Segments> segments = 1;
+}
+~~~~
+{: #figure-31 title="Control Service RPC API - Segment lookup.
+   This API is exposed on the SCION dataplane by the control
+   services of core ASes and exposed on the intra-domain protocol
+   network."}
```
nicorusti: @knmeynell could we use the same format as for the other section? 

- Service definition first
- then explanation of fields 
- then protobuf of SegmentsRequest and response

nicorusti:
@knmeynell Can you give it one last look and merge?

PR #150, opened 2025-10-17
Move Service Addressing from appendix to dedicated section

### LINKED ISSUE #109: Service addressing - rewrite and move to dedicated section
Service discovery was introduced in the draft in #15 & #24, however its description is not complete  

TODO: 

- [x] Move section to a dedicated paragraph
- [x] Update text so that it clearly answers questions form Alvaro
- [x] Service discovery now runs on UDP port 30252, we discussed in #7 why we do not specify it (because it is in the AS-local config). However, we need to write a bit more text about how each AS is doing the mapping
- [x] We should use proper example IPs instead of these
```    
// Supported address format for QUIC:
//  192.168.0.1:80
//  [2001:db8::1]:80
```
- [x] Add missing cross references from other sections (e.g. path lookup, registration, ?, etc...) 
- [x] Clarify that service addresses  are used to discover the CP service, but be mindful of not confusing it with what was called service discovery.
- [x] Explicitly indicate how the information is used.
- [x] Should we also move `Figure 21: Control Service gRPC API - Trust Material representation` to a dedicated section?

Feedback from  Alvaro Retana: 

>   How is the location of the Control Service discovered?  It is the main control plane component, but it is not clear to me how, for example, the non-core ASes know where to register, or who the PCBs should come from, etc.
  The Appendix (SCION Data Plane use by the SCION Control Plane) contains  information about it.  Consider "upgrading" the appendix to a section in the main text.

>> Yes, this is based ons service discovery that is currently described in the
> appendix. There is some text thad describes how service addresses are used 
> for control service discovery. It uses a reserved destination host address
> (0x0002), which gets then forwarded by the destination AS SCION border router
> to the right endpoint running the control service.
...

> I'm assuming that once the Control Service in an AS is found, the communication with it is direct (and not using the border router as a relay), right?
Also, the information from the discovery is used to establish the QUIC/H3 connection and authenticate that the PCBs originate from the correct source, right?

knmeynell:
Resolves #109 

<br class="Apple-interchange-newline">[SCION CP](https://scionassociation.github.io/scion-cp_I-D/Service-Addressing/draft-dekater-scion-controlplane.html)	[plain text](https://scionassociation.github.io/scion-cp_I-D/Service-Addressing/draft-dekater-scion-controlplane.txt)	[diff with main](https://author-tools.ietf.org/api/iddiff?url_1=https://scionassociation.github.io/scion-cp_I-D/draft-dekater-scion-controlplane.txt&url_2=https://scionassociation.github.io/scion-cp_I-D/Service-Addressing/draft-dekater-scion-controlplane.txt)

### DIFF
--- a/draft-dekater-scion-controlplane.md
+++ b/draft-dekater-scion-controlplane.md
@@ -2437,12 +2437,26 @@ Changes made to drafts since ISE submission. This section is to be removed befor
 
 Major changes:
 
+- New section "Distribution of Cryptographic Material" containing definitions formerly in the gRPC API appendix
+- New section "Destination Mapping" including a SIG reference
+- New section "Lookup Requests Message Format" containing definitions formerly in the gRPC API appendix
+- Move appendix "Use of the SCION Data Plane" to new section "Control Service Discovery"
 - Mention ConnectRPC as main RPC method instead of gRPC
-- Add small section about destination mapping
+- Remove appendix "Full Control Service gRPC API" and move corresponding protobuf definitions in new sections mentioned above
 
 Minor changes:
 
-- AS Entry Signature: fix order of terms in one formula
+- Rename Inter-ISD Beaconing into Core Beaconing for consistency
+- Clarify descriptions of fields in the `HeaderAndBody` message and that metadata must be empty
+- AS Entry Signature: fix order of terms in one formula, clarify validity and meaning of associated data
+- PCB Extensions: clarified text, added example of the `StaticInfoExtension` and informative reference
+- PCB Validity: clarify text on timestamp validity and time allowances
+- Reception of PCBs: mention that incoming link MUST be core or parent
+- PCB selection policies: discourage use for traffic engineering
+- Best PCBs Set Size: clarify tradeoffs and avoid normative language when unnecessary
+- Path reversibility: mention that destination endpoints should estimate MTU
+- Move considerations on SCMP Authentication to the security considerations section (Rogue SCMP Error Messages)
+- Security Properties: use normative language to clarify assumptions
 
 ## draft-dekater-scion-controlplane-09
 {:numbered="false"}


draft-dekater-scion-controlplane.md
```diff
@@ -429,7 +429,69 @@ The RPC messages are transported via {{Connect}}'s RPC protocol that carries mes
 
 In case of failure, RPC calls return an error as specified by the RPC framework. That is, a non-zero status code and an explanatory string.
 
-{{app-b}} provides details about the establishment of the underlying QUIC connections through the data plane.
+## Control Service Discovery
+
+The Control Plane RPC APIs rely on QUIC connections over UDP/SCION (see {{I-D.dekater-scion-dataplane}} which requires a connection to be initiated to identify the relevant peer (service resolution) and to select a path to it. Since the Control Service is itself the source of path segment information, the following bootstrapping processes apply:
```
nicorusti: > ... requires a connection to be initiated to identify the relevant peer .... 

This sentence is not really clear and IMHO too long. 

```suggestion
The Control Plane RPC APIs rely on QUIC connections over UDP/SCION (see {{I-D.dekater-scion-dataplane}}. Establishing such connection requires the initiator to identify the relevant peer (service resolution) and to select a path to it. Since the Control Service is itself the source of path segment information, the following bootstrapping processes apply:
```

The SCION Control Plane RPC APIs rely on QUIC connections carried by the SCION dataplane. The main difference between QUIC over native UDP and QUIC over UDP/SCION is the need for a UDP/SCION connection initiator to identify the relevant peer (service resolution) and to select a path to it. Since the Control Service is itself the source of path segment information, the following bootstrapping strategies apply:

nicorusti (changes requested):
Can you review my proposal and changes I did directly in the text?

nicorusti (approved):

nicorusti:
For the records, this is the ID diff
[Diff_ draft-dekater-scion-controlplane.txt - draft-dekater-scion-controlplane.txt.html](https://github.com/user-attachments/files/22991564/Diff_.draft-dekater-scion-controlplane.txt.-.draft-dekater-scion-controlplane.txt.html)

PR #156, opened 2025-10-30
Reorganised PCB message format section

### LINKED ISSUE #154: Brian Trammell's review CP
This document is the one most directly answering questions posed in RFC 9217, so I look forward to future pre-standardization work on the control plane to be discussed in PANRG.

## Technical Issues

- Section 2.3.4: I'm interested in the derivation of the parameters given here, in that there are a few system parameters that emerge from them that in turn dictate the performance and operational characteristics of a SCION internet. Were these derived from simulation, or from experience with the existing (much smaller than the big-I Internet) Anapaya and SCIONlab networks?

- Have the PCB set sizes been tested against realisitic reference interdomain topologies? The intra-AS (editorial: shouldn't this be inter-AS, since it's leafward?) PCB size seems reasonable in a transit-customer-heavy topology (like the relationships that IIUC SCION was originally designed for), but in a realistic peering-heavy topology with fairly decentralized peering points (cf. the internet in Europe), I could see the peering links dominate PCBs such that viable peering shortcuts would be lost at the leaves with only 50 available per PCB. 3.3.2 points to a small diameter assumption following Barabasi... but I'm not sure the dynamics driving that model are the same that lead to Internet interdomain connectivity.

- The core beacon size similarly seems to have a core AS model in mind (that of a small coterie of completely connected providers, analogous to "Tier 1" in the 20th century Internet topology) that may not apply in every ISD.

- Similarly, propagation delay is related to the availability risk of a link outage. The 60 second core beaconing interval implies that the maximum time that a disconnect event (i.e., links between one core AS and the others, or an inter-ISD link, being commpletely down) would lead to packet loss is 60 seconds; this is long enough that it would probably trigger application and transport layer session resets in the absence of widespread peering shortcuts (which the PCB size seems to limit, see above)

_KM - the above would all seem to be philosophical questions._ 

- [x] Section 7, general question: are there any rules for converting an ICMP message to SCMP, or vice versa? The protocol seems designed to support this, so I was missing a mention of it in this section.

## Editorial Issues

- [x] I found at least one missing Oxford comma with a clarity impact (Section 1 para 2 "and provide continuity".)
_KM - there are inconsistencies between US and British stylisation in various places (not all introduced by me!). I'm a proponent of the Oxford comma (Adrian may not be), but also 'e.g.' (British style) compared to 'e.g.,' (US style). For me, 'e.g.' uses less characters and looks cleaner, although I don't know what the RFC Editor style guide says about it._ 

- [x] Section 1.4.1: what about path segments between non-core ASs? I understand that coreward segments are up and leafward down, but that's not what this list says. --> we can say "There are no segments starting and ending at non-core ASes. IN these situations, several segments are combined during path lookup (REF to 5.1.  Lookup Process)" 

- [x] Section 1.5.3: is there any interoperability concern with respect to non-wire representations of ISD/AS numbers? I'm confused about the 2119 language here. Some explanatory text in the frontmatter of 1.5.3 about why we care about text representation would be useful.

- [x] Section 2.2.1: Figure 6, the top-down composition of a PCB, would be easier to read if it were broken up and interspersed through the text as a series of smaller, more focused diagrams.

- [x] Section 2.2.2.1 (and maybe elsewhere): "MUST consists".

knmeynell:
resolves #154

<br class="Apple-interchange-newline">[SCION CP](https://scionassociation.github.io/scion-cp_I-D/Reorganised-PCB-Message-Format/draft-dekater-scion-controlplane.html)	[plain text](https://scionassociation.github.io/scion-cp_I-D/Reorganised-PCB-Message-Format/draft-dekater-scion-controlplane.txt)	[diff with main](https://author-tools.ietf.org/api/iddiff?url_1=https://scionassociation.github.io/scion-cp_I-D/draft-dekater-scion-controlplane.txt&url_2=https://scionassociation.github.io/scion-cp_I-D/Reorganised-PCB-Message-Format/draft-dekater-scion-controlplane.txt)

### DIFF
--- a/draft-dekater-scion-controlplane.md
+++ b/draft-dekater-scion-controlplane.md
@@ -638,7 +638,7 @@ Path Segment 4 |             |     |             |     |             |
 {: #figure-4 title="Possible up- or down segments for AS Z"}
 
 
-## Path-Segment Construction Beacons (PCBs) Message Format {#pcbs}
+## PCB Message Format {#pcbs}
 
 Each PCB is comprised of a message containing the following top-level fields:
 
@@ -906,7 +906,7 @@ The following code block defines the signed body of one AS entry in Protobuf mes
 - `mtu`: The maximum transmission unit (MTU) that is supported by all intra-domain links within the current AS. This value is set by the control service when adding the AS entry to the beacon. How the control service obtains this information is implementation dependent. Current practice is to make it a configuration item.
 - `extensions`: List of (signed) extensions (optional). PCB extensions defined here are part of the signed AS entry. This field SHOULD therefore only contain extensions that include important metadata for which cryptographic protection is required. For more information on PCB extensions, see [](#pcb-ext).
 
-##### Hop Entry {#hopentry}
+#### Hop Entry {#hopentry}
 
 ~~~
 
@@ -941,7 +941,7 @@ The following code block defines the hop entry component `HopEntry` in Protobuf 
 
 In this description, MTU and packet size are to be understood in the same sense as in {{RFC1122}}. That is, exclusive of any layer 2 framing or packet encapsulation (for links using an underlay network).
 
-##### Peer Entry {#peerentry}
+#### Peer Entry {#peerentry}
 
 ~~~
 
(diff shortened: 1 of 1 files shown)


nicorusti (approved):
I removed one level of indentation and reduced a bit other nested indentations that are IMHO a bit hard to follow. 
LGTM, @knmeynell can you have one last look and merge?

PR #163, opened 2025-11-11
Security considerations: add section on Attacks on time sources

### LINKED ISSUE #160: Security considerations: Attacks to time sources
From Eliot: 
> In the dataplane draft, I think you need to add a security consideration relating to attacks on time sources.  If you're using NTP and NTP is attacked, what are the implications?  Same with GPS.  That should happen before this goes to the IESG.

We currently have sections about the impact of clock drift (introduced in https://github.com/scionassociation/scion-cp_I-D/issues/18  ) but I don'm think we say much on the security aspect of it. This issue might also affect -controlplane.

[Slack discussion](https://scionproto.slack.com/archives/C8ADA9CEP/p1762452155817339)

Relates to past issues: 
- #18
- #117

nicorusti:
Resolves #160

### DIFF
--- a/draft-dekater-scion-controlplane.md
+++ b/draft-dekater-scion-controlplane.md
@@ -2150,7 +2150,7 @@ Care should be taken to maintain coarse time synchronization among Control Servi
 
 - A control service instance: its beaconing process may halt as it cannot verify the validity of received PCBs (see [](#pcb-validity)) or correctly add timestamps to propagated PCBs (see [](#pcb-appending)).
 - An endpoint: the endpoint may fail to verify path segments during path lookup.
-- A router: packets may be dropped if the time offset exceeds the Hop Field expiration time (between 337.5s and a day, see [](#hopfield)).
+- A router: packets may be dropped ahead of the control service intended expiration time (see [](#hopfield)).
 
 ## Denial of Service Attacks {#dos-cp}
 


knmeynell (approved):

draft-dekater-scion-controlplane.md
```diff
@@ -2143,6 +2144,14 @@ To defend against this attack, methods to detect the wormhole attack are needed.
...
 Note that this would be mitigated through authentication of SCMP messages. Authentication is not specified here since it is currently still experimental.
 
+## Attacks on time sources {#time-security}
+
+Care should be taken to maintain coarse time synchronization among Control Service instances and other system components, as discussed in [](#clock-inaccuracy). An adversary that significantly alters the system time of a component can disrupt SCION operations:
+
+- A control service instance: its beaconing process may halt as it cannot verify the validity of received PCBs (see [](#pcb-validity)) or correctly add timestamps to propagated PCBs (see [](#pcb-appending)).
+- An endpoint: the endpoint may fail to verify path segments during path lookup.
+- A router: packets may be dropped if the time offset exceeds the Hop Field expiration time (between 337.5s and a day, see [](#hopfield)).
```
shitz: The time offset doesn't need to exceed the hopfield's expiration time...

If the clock in the router is N seconds ahead of the control service's clock then the packet at the router will be dropped N seconds before the control service's "intended expiration".

knmeynell (approved):

PR #164, opened 2025-11-12
Clarifies validation of path segments

### LINKED ISSUE #162: Path lookup: clarify validation of segments by endpoints
Follow up to #160.  I think we are missing some details on segment validation by endpoints in the path lookup process, in between [these two steps](https://www.ietf.org/archive/id/draft-dekater-scion-controlplane-11.html#section-5.1-5.4.1): 
```
4. The Control Service of the source AS returns all retrieved path segments to the source endpoint.
5. Once it has obtained all path segments, the source endpoint combines them into an end-to-end path in the data plane.
```

Does the endpoint, between 4 and 5: 
- verify that the the timestamp in `segment_info` is not too late?  (ref to  [](#seginfo)
- verify that the AS entry signature?  (reference to #sign)
- request any missing intermediate certificates to the CP if they are not already available? (Ref to #crypto-api)

Slack [thread](https://scionproto.slack.com/archives/C8ADA9CEP/p1762877541765009)

knmeynell:
Resolves #162 

[Diff with main](https://author-tools.ietf.org/api/iddiff?url_1=https://scionassociation.github.io/scion-cp_I-D/draft-dekater-scion-controlplane.txt&url_2=https://scionassociation.github.io/scion-cp_I-D/Clarifies-validation-of-path-segments/draft-dekater-scion-controlplane.txt)


nicorusti (approved):
LGTM, waiting for a double check by devs and/or Sam

Issue #182, opened 2026-01-26
SCMP errors 1 and 4?


tzaeschke:
Table 5 in section 7.2 does not list the SCMP errors 1 (destination unreachable) and 4 (parameter problem) as they are defined here: https://docs.scion.org/en/latest/protocols/scmp.html#scmp-error-messages

https://github.com/scionassociation/scion-cp_I-D/blob/02c446ba36c25d3031808f1bd1a4d43fb274ee12/draft-dekater-scion-controlplane.md?plain=1#L1767-L1774

Have these been removed?

nicorusti:
I was going through the comments of the ISE editor in #177 . One of them is: 
> 16. Section 2.2.3 what if a receiver doesn't understand an extension?
 
And the answer is: it MAY send a SCMP parameter problem. I will consult him about the two missing SCMP messages.

PR #184, opened 2026-01-27
SCMP specification

tzaeschke:
This PR removes unnecessary entries from the SCMP specification section. See also #182


nicorusti (commented):
Le'ts keep this on hold until I get feedback from the ISE Editor about the two missing SCMP messages

PR #185, opened 2026-01-28
Simplified figures

### LINKED ISSUE #188: 10. There are small issues with Figures 3 and 4 with side arrows not being properly represented in SVG in the HTML version.  I imagine you are pressed for space in the ASCII version?  If you I could ask one of the aasvg authors to have a look.  Before I do that, however, I am not sure those arrows are appropriate.  I think you're just meaning to expand what the PCB in each case is, rather than show flow of communication.  In that case, I might remove the arrows.

knmeynell:
Resolves #188 
[Diff with main](https://author-tools.ietf.org/api/iddiff?url_1=https://scionassociation.github.io/scion-cp_I-D/draft-dekater-scion-controlplane.txt&url_2=https://scionassociation.github.io/scion-cp_I-D/Redrawn-Figures-2-4/draft-dekater-scion-controlplane.txt)
I've reduced PCBs to one single box each, to avoid problems with confusing arrows.

### DIFF
--- a/draft-dekater-scion-controlplane.md
+++ b/draft-dekater-scion-controlplane.md
@@ -535,7 +535,7 @@ AS Y also has two peering links to its neighboring peers V and W, through the in
                            +----+---+----+
                            |    AS Z     |
 ~~~
-{: #figure-3b titl	e="Intra-ISD PCB propagation from the ISD core to child ASes - Part 2"}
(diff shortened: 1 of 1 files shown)


nicorusti (approved):

nicorusti (approved):
@knmeynell I made some more changes, what do you think? Can you 4 eye check, and if they look ok for you, can you merge? 

Image

Image

nicorusti (approved):
Image

Image

PR #192, opened 2026-01-29
attacks to time sources: recommend secure time sync

### LINKED ISSUE #160: Security considerations: Attacks to time sources

nicorusti:
Resolves #160
[Diff with main](https://author-tools.ietf.org/api/iddiff?url_1=https://scionassociation.github.io/scion-cp_I-D/draft-dekater-scion-controlplane.txt&url_2=https://scionassociation.github.io/scion-cp_I-D/time_security/draft-dekater-scion-controlplane.txt)

Slack [question](https://scionproto.slack.com/archives/C8ADA9CEP/p1769710277323639?thread_ts=1762452155.817339&cid=C8ADA9CEP)

### DIFF
--- a/draft-dekater-scion-controlplane.md
+++ b/draft-dekater-scion-controlplane.md
@@ -59,6 +59,7 @@ normative:
     target: https://pubs.opengroup.org/onlinepubs/9799919799/basedefs/V1_chap04.html
 
 informative:
+  BCP223:
   ISD-AS-assignments:
     title: "SCION Registry"
     date: 2026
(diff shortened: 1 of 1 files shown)


nicorusti:
@knmeynell please review but wait before merging. I want to hear back on Slack

PR #193, opened 2026-01-29
path lookup process: reword to clarify that requests are in parallel

### LINKED ISSUE #183: Segment request procedure seems wrong
The segment request procedure in section 5.1 seem wrong to me:

https://github.com/scionassociation/scion-cp_I-D/blob/02c446ba36c25d3031808f1bd1a4d43fb274ee12/draft-dekater-scion-controlplane.md?plain=1#L1554-L1560

It also seems to contradict the description in section 5.1.1:

https://github.com/scionassociation/scion-cp_I-D/blob/02c446ba36c25d3031808f1bd1a4d43fb274ee12/draft-dekater-scion-controlplane.md?plain=1#L1571-L1577

**Note:** To avoid misunderstanding, I assume the [New endhost API](https://github.com/scionproto/scion/issues/4834) is not yet standardized and therefore not described here. If this indeed describes the new endhost API, then Section 5.1.1 is wrong.

Specifically:

- Ad 1. `The Control Service has up segments stored in its path database and additionally checks if it has appropriate core segments and down segments stored as well, and in this case returns them immediately in a SegmentsResponse.`
This sounds like as if, when receiving an UP segment request, the CS would also check for CORE and DOWN segments (and return them as well).
I don't this is not the case. As described in section 5.1.1, requests for UP, CORE and DOWN are issued and answered separately.

- Ad 3. `The Control Service of the source AS combines up segments with the newly retrieved core segments. The Control Service then queries the Control Services of the remote core ASes in the destination ISD to fetch down segments to the destination AS. To reach the remote core ASes, the Control Service of the source AS uses the previously obtained and combined up segments and core segments.`
I don't think that is the case, the CS does not combine path segments.

I wonder if this could be changed to:

```
1. The source endpoint queries the Control Service in its own AS (i.e. the source AS) for the required segments by sending a `SegmentsRequest` separately for up, core and down segments. 
   1. The Control Service has up segments stored in its path database and returns them immediately in a `SegmentsResponse`. 
   2. The Control Service may have appropriate core segments available in its cache, and in this case returns them immediately in a `SegmentsResponse`. If there are no appropriate core segments, the Control Service in the source AS queries the Control Services of the reachable core ASes in the source ISD for core segments to core ASes in the destination ISD. To reach the core Control Services, the Control Service of the source AS uses the locally stored up segments. 
   3. The Control Service may have appropriate down segments available in its cache, and in this case returns them immediately in a `SegmentsResponse`. If there are no appropriate down segments, the Control Service of the source AS queries the Control Services of the remote core ASes in the destination ISD to fetch down segments to the destination AS. To reach the remote core ASes, the Control Service of the source AS uses the previously obtained and combined up segments and core segments. 
2. The Control Service of the source AS returns all retrieved segments to the source endpoint. 
3. As the source endpoint receives each path segment, it verifies the `SegmentInformation` timestamp validity (see [](#pcb-validity)), the AS entry signature for each AS entry (see [](#sign)), and requests any missing AS or intermediate certificates from the Control Service (see [](#crypto-api)). 
4. Once it has obtained some valid path segments, the source endpoint combines them into an end-to-end path in the data plane. 
5. The destination endpoint, once it receives the first packet, MAY reverse the path in the received packet in order to construct a response. This ensures that traffic flows on the same path bidirectionally. 
```

Alternatively, I wonder if these points could all be removed. The endhost activity is already described in section 5.1.1. CS activity is described in section 5.2. And the last point about path reversal is described elsewhere in the document.
I suggest removing this list.

Path segment combination is also described in the [DP draft](https://github.com/scionassociation/scion-dp_I-D/blob/main/draft-dekater-scion-dataplane.md#path-construction-segment-combinations-construction)

nicorusti:
Resolves #183 

[Diff with main](https://author-tools.ietf.org/api/iddiff?url_1=https://scionassociation.github.io/scion-cp_I-D/draft-dekater-scion-controlplane.txt&url_2=https://scionassociation.github.io/scion-cp_I-D/path_lookup_fix/draft-dekater-scion-controlplane.txt)
[wdiff](https://author-tools.ietf.org/diff?url_1=https://scionassociation.github.io/scion-cp_I-D/draft-dekater-scion-controlplane.txt&url_2=https://scionassociation.github.io/scion-cp_I-D/path_lookup_fix/draft-dekater-scion-controlplane.txt&wdiff=1)

### DIFF
--- a/draft-dekater-scion-controlplane.md
+++ b/draft-dekater-scion-controlplane.md
@@ -2370,7 +2370,7 @@ Changes made to drafts since ISE submission. This section is to be removed befor
 ## draft-dekater-scion-controlplane-15
 {:numbered="false"}
 
-- Lookup Process: reword steps to clarify that requests happen in parallel
+- Lookup Process: reword steps to clarify how an endpoint requests path segments
 - Wording polish following ISE Editor's feedback
 - Figures 2, 3, 4: improve arrows in SVG version
 - Remove redundant section 1.7. Resistance to partitioning


nicorusti:
Waiting for @tzaeschke 's feedback before merging

draft-dekater-scion-controlplane.md
```diff
@@ -1442,10 +1442,12 @@ The actual number of required path segments depends on the location of the desti
...
-4. The Control Service of the source AS returns all retrieved path segments to the source endpoint.
+1. The source endpoint queries the Control Service in its own AS (i.e. the source AS) for the required segments by sending up to three parallel `SegmentsRequest`, respectively for up, core and down segments.
+2. The Control Service of the source AS returns segments to the endpoint in up to three `SegmentsResponse`. Specifically, for each segment type:
+
+   - up segments are stored in its path database. It returns them immediately.
+   - core segments may be cached. Otherwise, it queries the Control Services of the reachable core ASes in the source ISD for core segments to core ASes in the destination ISD. To reach the core Control Services, the Control Service of the source AS uses the locally stored up segments. Once obtained, it returns the core segments.
```
tzaeschke:
```suggestion
   - core segments may be cached in the local Control Service. Otherwise, the local Control Service queries the Control Services of the reachable core ASes in the source ISD for core segments to core ASes in the destination ISD. To reach the core Control Services, the Control Service of the source AS uses the locally stored up segments. Once obtained, it returns the core segments.
```

tzaeschke: I did this to make clear where it is cached, that may not be clear otherwise?

draft-dekater-scion-controlplane.md
```diff
@@ -1442,10 +1442,12 @@ The actual number of required path segments depends on the location of the desti
...
+   - down segments may be cached. Otherwise it queries the Control Services of the remote core ASes in the destination ISD to fetch down segments to the destination AS. To reach the remote core ASes, the Control Service of the source AS uses the previously obtained and combined up and core segments. Once obtained, it returns the down segments.
```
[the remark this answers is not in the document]
tzaeschke: Oh it is merged already. Nevermind.

PR #203, opened 2026-02-03
Registration: clarify which  core AS registers down segments

### LINKED ISSUE #202: Segment registration - inconsistency
We say in [section 4](https://www.ietf.org/archive/id/draft-dekater-scion-controlplane-14.html#section-4-3): 
> As a result, a core AS's Control Service contains all intra-ISD path segments registered by the non-core ASes of its ISD. 

But then in [4.3](https://www.ietf.org/archive/id/draft-dekater-scion-controlplane-14.html#section-4.3-1) we say: 
> The Control Service of a non-core AS has to register the newly created down segments with the Control Services of the core ASes that originated the corresponding PCBs. 

This was developed in https://github.com/scionproto/scion/pull/3747 and discussed [in Slack](https://scionproto.slack.com/archives/C8ADA9CEP/p1588670470176500) in 2020.

Wording in section 4 comes from [version -01](https://github.com/scionassociation/scion-cp_I-D/commit/38e5e307f15a9371a4b7a9ca14cd9071a942424f) (August 2023) and this is likely how it used to be in an early design. Docs [was updated](https://github.com/scionproto/scion/pull/4361/changes#diff-598a1a36ae29697fdd68e36f6dd1ba3b3367c490d3c6c22329f0e3c9b4a978f8R193) a few weeks later (23 Aug 2023) with a different wording.

nicorusti:
Resolves #202

### DIFF
--- a/draft-dekater-scion-controlplane.md
+++ b/draft-dekater-scion-controlplane.md
@@ -1371,7 +1371,7 @@ Every registration period, the Control Service of a non-core AS performs the fol
 
 1. The Control Service selects the PCBs that it wants to transform into down segments from the candidate PCBs in the Beacon Store.
 2. The Control Service "terminates" the selected PCBs by performing the steps described in [](#term-pcb). From this moment on, the modified PCBs are called **down segments**.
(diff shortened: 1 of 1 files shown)


nicorusti:
@knmeynell can you review, but please wait for confirmation from this [Slack](https://scionproto.slack.com/archives/C8ADA9CEP/p1770153381067049) thread before merging

PR #209, opened 2026-02-06
Chapter 2 review

### LINKED ISSUE #219: Core beaconing: ensure consistent wording
We use different wording about how core ASes create PCBs: 

initiating

- Control Service of a core AS either initiates PCBs or

originat*

- This means that the originating core AS MUST add its AS entry to each PCB it creates, 
- timestamp: The timestamp indicates the creation time of this PCB. It is set by the originating core AS
- and many more 

creat*

- For this, the Control Services of core ASes create PCBs
- The timestamp indicates the creation time of this
- The core beaconing process creates path segments f
- This duration is relative to the PCB creation timestamp
- The core beaconing process creates path segments
- The Control Service adds the newly created core segments 
- The Control Service of a non-core AS has to register the newly created down segments 

Note that the RPC is called `SegmentCreationService`, so there is a plus in using creation

I think a potential approach is to leave "creat*" for both core and non-core AS, and be more specific when talking about how Core ASes start beaconing by using originat*

tzaeschke:
Another pass (more detailed).
Thanks to @FR4NK-W for discussion on this one.

[Diff with main](https://author-tools.ietf.org/api/iddiff?url_1=https://scionassociation.github.io/scion-cp_I-D/draft-dekater-scion-controlplane.txt&url_2=https://scionassociation.github.io/scion-cp_I-D/tz-section-2/draft-dekater-scion-controlplane.txt)
Resolves #219

### DIFF
--- a/draft-dekater-scion-controlplane.md
+++ b/draft-dekater-scion-controlplane.md
@@ -145,39 +145,6 @@ informative:
       -
         ins: SCION Association
         org: SCION Association
-  SCIONLAB:
-    title: SCIONLAB - A Next-Generation Internet Testbed
-    date: 2020
-    target: https://ieeexplore.ieee.org/abstract/document/9259355
-    author:
-      -
-        ins: J. Kown
-        name: Jonghoon Kwon
-        org: ETH Zuerich
-      -
-        ins: J. García-Pardo
-        name: Juan A. García-Pardo
-        org: ETH Zuerich
-      -
-        ins: M. Legner
-        name: Markus Legner
-        org: ETH Zuerich
-      -
-        ins: F. Wirz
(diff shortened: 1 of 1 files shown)


draft-dekater-scion-controlplane.md
nicorusti:
```suggestion
- propagates each selected PCB to the selected egress interface(s) associated with it. Selection policies are described in [](#path-segment-prop).
```

Since we removed it a few lines above, I'd be more explicit here

nicorusti: GitHub shows the diff weird. All I did was to rename 
> This is described  

With 
> Selection policies are described

PR #214, opened 2026-02-09
Chapter 8 Review

tzaeschke:


draft-dekater-scion-controlplane.md
```diff
@@ -2103,7 +2103,7 @@ A core AS may reach other core ASes in the same ISD via other ISDs, depending on
 
 ## Manipulation of the Beaconing Process by a Core Adversary {#topdown-manipulate}
 
-The first risk to the beaconing process comes from an adversary controlling one or more core ASes in an ISD. If the adversary stops all core AS(es) within an ISD from propagating PCBs, the discovery of new paths will halt. In this case, downstream ASes will notice that PCBs are no longer being propagated, but all previously discovered and still valid paths remain usable for data plane forwarding until they expire. This is an unlikely scenario, as it would require compromise of all core ASes within an ISD.
```
tzaeschke: The key problem here is that we don´t need _all_ core ASes to be compromised, only those that pertain to a given downstream AS.

draft-dekater-scion-controlplane.md
```diff
@@ -2103,7 +2103,7 @@ A core AS may reach other core ASes in the same ISD via other ISDs, depending on
...
+The first risk to the beaconing process comes from an adversary controlling one or more core ASes in an ISD. Non-core ASes depend on core ASes for beaconing. If an adversary stops a core AS from propagating PCBs, the discovery of new paths for will halt. In this case, downstream ASes will notice that PCBs are no longer being propagated, but all previously discovered and still valid paths remain usable for data plane forwarding until they expire. Since non-core ASes are often connected to multiple core ASes, they will continue to receive beacons from other core ASes. This is an unlikely attack scenario, as it would require compromise of all core ASes of a given downstream AS to be effective.
```
nicorusti: How about this wording: 

```suggestion
The first risk to the beaconing process comes from an adversary controlling one or more core ASes in an ISD. Non-core ASes depend on core ASes for beaconing. If an adversary compromises all upstream core ASes of a target AS and suppresses PCB propagation, the discovery of new paths for will halt. Although the downstream AS will no longer receive new PCBs, previously discovered and still valid paths remain usable for data plane forwarding until they expire. This is an unlikely attack scenario, as it would require compromise of all core ASes of a given downstream AS to be effective.
```

Issue #213, opened 2026-02-09
Clarify SCMP processing rules


tzaeschke:
Section 6.4:
```
* If an SCMP error message of unknown type is received at its destination, it MUST be passed to the upper-layer process that originated the packet that caused the error, if it can be identified.

* If an SCMP informational message of unknown type is received, it MUST be silently dropped.

* [...]

* In case the implementation is required to pass an SCMP error message to the upper-layer process, the upper-layer protocol type is extracted from the original packet in the body of the SCMP error message and used to select the appropriate process to handle the error. In case the upper-layer protocol type cannot be extracted from the SCMP error message body, the SCMP message MUST be silently dropped.

* [...]
```

This could benefit from some clarification:

- Ad 1): How can a socket determine if an incoming packet was triggered by an outgoing packet (let alone which process)? Note: the type is unknown which means the header length is unknown so the payload cannot be extracted. Even with payload available, it is unclear how that would work.
- Ad 1): What is the purpose of forwarding unknown SCMP messages to the application layer? Even for *known* SCMP messages it is not obvious how this is meant to work, should processes know SCMP messages? And for *unknown* SCMP messages? 
- Why are unknown types of Error and Info treated differently?
- ~Shouldn't there be a process requirement/recommendation about authenticating SCMP messages before processing?~ EDIT:  Section 8.3 states that SCMP authentication is still experimental and not specified here.
- Ad 4): How does a SCION library know whether it is required to do that?

A partial insight: Point 4) suggests that the upper layer protocol is used to identify upper layer processes. Does this refer to a design where a SCION kernel socket passes traffic on to a UDP kernel socket (here called a "process")?

See also discussion here: https://scionproto.slack.com/archives/C8ADA9CEP/p1770398184282579

tzaeschke:
My preferred solution would be to remove rules 1), 2), and 4), and keep only 3) and 5). 

If that is too big of a change (so late in the process), #217 offers a small(-ish) change to clarify the applicability of the rules.

nicorusti:
P.S. This text mimics [rfc4443 section-2.4](https://datatracker.ietf.org/doc/html/rfc4443#section-2.4 ) 

```
Implementations MUST observe the following rules when processing
   ICMPv6 messages (from [[RFC-1122](https://datatracker.ietf.org/doc/html/rfc1122)]):

   (a) If an ICMPv6 error message of unknown type is received at its
       destination, it MUST be passed to the upper-layer process that
[...]

PR #217, opened 2026-02-09
Clarify SCMP processing

tzaeschke:
This aims to solve #213.

My preferred solution would be to remove rules 1), 2), and 4), and keep only 3) and 5).
If that is too big of a change (so late in the process), this PR offers a smallish change to clarify the applicability of the rules.


nicorusti:
I proposed some new wording removing the offending lines Til pointed to, but I am still not convinced. I posted some thoughts in the issue

nicorusti:
@tzaeschke I reverted this, let's continue doing changes on the same branch in #223

